PR #101, opened 2026-06-15
fix(qlik): hard-fail discovery on empty/inaccessible load script (silent tables=0)

twells89:
## What was happening
While reproducing a Qlik **Sales Analysis / Executive Dashboard** migration for a quickstart writeup, the migration "wasn't doing it well." Root cause: `qlik-discover.py` fetched the app load script — **the data-model source of truth** — via `qlik(..., parse_json=False)`, whose non-zero-exit WARN was gated on `parse_json` being **true**. So a failed `GetScript` (e.g. `GENERIC ACCESS DENIED` when the connecting identity doesn't own / can't read the app) was swallowed *entirely*:

- `script.qvs` written as **0 bytes**
- `parse_script("")` → **tables=0**
- discovery printed a cheerful `tables=0 … Next: run migrate-qlik.rb`
- the converter then built a Sigma data model **with no data** — no clue why.

## Fix
- `qlik()` now warns on **any** non-zero exit (incl. `parse_json=False` calls).
- `_script()` captures the GetScript error text.
- Discovery now **hard-fails (exit 3)** when the script is empty AND the app isn't DirectQuery, printing the GetScript error + three concrete fixes (run as owner / copy-own the app / publish to a readable space). DirectQuery apps are exempt.
- QUICKSTART "Reference & gotchas" documents this as the #1 cause of an empty Qlik migration.

## Validation (live)
Run against a real access-restricted app (Executive Dashboard, owned by another user):
```
FATAL: load script is empty — cannot build a data model (tables=0).
  GetScript failed: ... GetScript: Access denied (5 GENERIC ACCESS DENIED)
  ... Fix one of: (a) run discovery as the app OWNER, (b) copy/transfer ... (c) publish ...
```
exit 3 — instead of silently emitting `tables=0`. `py_compile` clean.

### DIFF
--- a/plugins/cognos-to-sigma/skills/cognos-to-sigma/SKILL.md
+++ b/plugins/cognos-to-sigma/skills/cognos-to-sigma/SKILL.md
@@ -213,6 +213,11 @@ the workbook and re-runs the error-column gate. **`apply-layout.mjs` then gives 
 clean 24-col grid** (controls on top, content stacked full-width with per-kind heights) —
 Sigma auto-arrange otherwise squishes every element to the same height. It writes the
 top-level `spec.layout` XML (matched to readback ids) and confirms it survives readback.
+**It then runs the shared layout-quality lint as a gate** (`scripts/lib/layout_lint.rb`,
+vendored byte-identical; cognos shells out to ruby) on the final readback spec — flags
+raw-id element display names, controls orphaned outside containers, and generic Sigma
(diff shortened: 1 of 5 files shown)


twells89:
Superseded. Split for clarity:

- **qlik discovery hard-fail** (the stated fix) → landed clean as #147 (merged), 2 files, no unrelated changes.
- **cognos layout-lint gate** that was bundled here → already on `main` via #100 (merged 2026-06-15); this branch predated that merge so it re-contained the same commit.

Closing in favor of #147 + #100. No work lost.

PR #166, opened 2026-06-24
looker-to-sigma: carry bar orientation + grid data bars through the converter

twells89:
## Why

Migrating Looker dashboard 11 "Retail Performance Overview" surfaced two **render-affecting** features the converter silently dropped from the workbook. Both are now carried through, with a committed regression.

## What

**1. Bar orientation (`looker_bar` → horizontal)**
Looker `looker_bar` renders horizontal bars and `looker_column` vertical, but both mapped to a vertical Sigma `bar-chart`. `build_workbook.py` now emits `orientation: horizontal` for `looker_bar` and omits the key for `looker_column` (Sigma's vertical default). Field verified: `sigma-workbooks` `charts.md`.

**2. Grid data bars (`series_cell_visualizations` → `conditionalFormats: dataBars`)**
Looker grids draw in-cell bars on a measure column, colored by value (low→high gradient).
- `fetch_looker_dashboard.py` + `parse_lookml_dashboard.py` capture `vis_config.series_cell_visualizations` into the contract as `cellVisualizations: {field: {scheme}}`.
- `build_workbook.py` emits element-level `conditionalFormats: [{type: dataBars, columnIds:[<calc col>], scheme?}]` on the table. Shape verified: `sigma-workbooks` `tables.md`.

**Render-only caveat (documented, not silently dropped):** Looker frequently does **not** return `series_cell_visualizations` from the dashboard/query API even when the rendered dashboard shows the bars (observed on dash 11 — confirmed empty via query, result_maker, and dashboard_element endpoints). That case can't be auto-detected from the contract, so it's documented as a Phase-4a visual-QA manual-add step in `SKILL.md`.

## Tests

`tests/test_grid_fidelity.py` — extractor unit tests (`_cell_viz` / `norm_cell_viz`) + end-to-end build assertions through the offline pipeline (flips the fixture column tile to `looker_bar`, adds cell-viz to the grid tile, asserts orientation + dataBars; also asserts the unmodified fixture stays vertical with no dataBars). `python3 tests/test_grid_fidelity.py` → ALL PASS.

## Verified live

Both fixes were applied to the migrated workbook `dd3e3e0c` (Conversion testing) and confirmed in the Sigma render against the Looker source: Category bar now horizontal, Units Ordered / Total Quantity column now shows value-gradient data bars.

### DIFF
--- a/plugins/looker-to-sigma/skills/looker-to-sigma/SKILL.md
+++ b/plugins/looker-to-sigma/skills/looker-to-sigma/SKILL.md
@@ -623,17 +623,24 @@ Newspaper layout math (a single arithmetic transform, no spatial heuristic):
   both map to a Sigma `bar-chart`. `build_workbook.py` emits `orientation: horizontal` for
   `looker_bar` and omits the key (Sigma's vertical default) for `looker_column`. Field verified:
   `sigma-workbooks` `charts.md`.
-- **Grid data bars (`series_cell_visualizations`).** A Looker grid can draw in-cell bars on a
-  measure column, colored by VALUE (low→high gradient). When the tile's `vis_config` exposes
-  `series_cell_visualizations`, `fetch_looker_dashboard.py` captures it (contract
-  `cellVisualizations: {field: {scheme}}`) and `build_workbook.py` emits an element-level
-  `conditionalFormats: [{type: dataBars, columnIds:[<calc col>], scheme?}]` on the table (verified
-  spec shape: `sigma-workbooks` `tables.md`). **Render-only caveat:** Looker frequently does **not**
-  return `series_cell_visualizations` from the dashboard/query API even when the rendered dashboard
-  shows the bars (observed on a fixture-built UDD) — that case can't be auto-detected from the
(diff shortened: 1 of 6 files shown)


twells89:
Update after live verification of Sigma data-bar semantics:

**Sigma data bars are sign-colored, not a value gradient.** The `Format rule → Data bars` UI exposes only a *Negative color* + *Positive color*; the bar fill is chosen by sign and can't vary by value. Probing confirmed it — a 3-stop `scheme` collapses to the single positive color (every all-positive bar rendered the middle stop), and `domain` had no effect on color.

So Looker's **value-colored** grid bars don't map to `dataBars` — they map to a Sigma **Color scale** (`backgroundScale`, cell tint low→high). Latest commit changes `build_workbook.py` to emit:
- `backgroundScale` (palette → scheme) + a warning when Looker declared a value palette (`custom_colors`);
- `dataBars` (plain magnitude, no scheme) otherwise.

Regression updated to assert both branches; SKILL.md Phase 3b corrected. Verified live on wb `dd3e3e0c`: the Color scale renders a real pink→purple→blue spread (cell hue 214–320°) tracking value.

PR #178, opened 2026-06-25
tableau-to-sigma: fix EDNA crosstab / filter / parameter coverage gaps

twells89:
## Why

A customer's EDNA **Partner Landscape** workbook (10 MB, 86 worksheets, 8 dashboards) migrated poorly: grouped/pivot tables came out as flat tables, most dashboard filters were missing, and the measure-switcher parameters "weren't handled well." We don't have EDNA's warehouse data, so this re-creates the *shapes* on the live CSA.TJ retail star and pins them as a regression fixture.

Diagnosed three regressions on the real `.twb`, each a one-spot root cause:

| # | Symptom | Root cause | EDNA before → after |
|---|---|---|---|
| 1 | Grouped/pivot tables built as flat tables | `parse-twb-layout` `is_crosstab` required a `Text`/`Square` mark, but the tables use Tableau's **default `Automatic`** mark (renders Measure-Names as a text grid) | crosstabs detected **2 → 44** |
| 2 | ~40 dashboard quick-filters dropped | `COL_BY_GUID` only registered caption-bearing columns + truncated multi-word names to the first word; `guid_from_param` choked on `(group)`/`(copy)`/slash | shared filters resolved **0/70 → 70/70** |
| 3 | Measure-switcher params showed raw `1/2/3` | `build-charts` set `labels: []` on the segmented control | now shows alias labels (TCV/ACV/…); params deduped **604 → 38** |

Plus a `preflight_lint` C2 false-positive: it flagged the `{kind:manual}` segmented source (which Sigma **accepts** — live-verified) as malformed, which would have blocked every measure-switcher at the migrate gate. Fixed in the canonical `shared/lib` copy and synced to all 8 converters.

## Validation

- **Live E2E on `CSA.TJ.ORDER_FACT`** (org tj-wells-1989): POSTed DM + workbook, applied layout, rendered the page. The pivot renders as a true grouped crosstab with a **Grand-total row + column**; the segmented control shows the alias labels; flipping it Net Revenue → Quantity swaps the measure column (grand total **119,038 → 1,593**) while the static Net Revenue column holds — proving the parameter switches the calculation. Throwaway artifacts deleted.
- **New corpus fixture** `tableau/partner-crosstab-controls` pins all four fixes through `parse-twb-layout` + `build-charts` (exercises caption-less, group, and slash filter columns + the integer measure-swap param). `corpus --check`: **4/4 tableau cases pass** (no regressions).

### DIFF
--- a/corpus/tableau/partner-crosstab-controls/MANIFEST.md
+++ b/corpus/tableau/partner-crosstab-controls/MANIFEST.md
@@ -32,7 +32,7 @@ Synthetic XML (no live tenant), modeled on real Tableau 2024.1 `.twb` shapes.
 
 | File | What it is |
 |---|---|
-| `workbook-content.twb` | Synthetic workbook: 1 Automatic-mark crosstab worksheet `Metric Partner Closed Won` (Partner Name rows × Created FYQQ + Measure Names cols), 1 dashboard, a `<shared-view>` with 3 quick filters exercising all the failure modes — caption-less multi-word (`Customer Geo`), a Tableau GROUP (`Partner Level (group)`), and a slash name (`Mkt Sourced/Influenced`) — plus an integer measure-swap parameter (`Summary $ Choose`, 1→TCV / 2→Product TCV / 3→ACV) with value aliases |
+| `workbook-content.twb` | Synthetic workbook: 1 Automatic-mark crosstab worksheet `Metric Partner Closed Won` (Partner Name rows × Created FYQQ + Measure Names cols), 1 Automatic-mark `Bookings by Quarter Bar` worksheet that is the OVER-FIRE GUARD (Measure Names on cols but the measure VALUES on rows as `[Multiple Values]` → a bar, must NOT be a crosstab), 1 dashboard, a `<shared-view>` with 3 quick filters exercising all the failure modes — caption-less multi-word (`Customer Geo`), a Tableau GROUP (`Partner Level (group)`), and a slash name (`Mkt Sourced/Influenced`) — plus an integer measure-swap parameter (`Summary $ Choose`, 1→TCV / 2→Product TCV / 3→ACV) with value aliases |
 | `get-workbook.json`, `master-columns.json` | Minimal discovery fixtures so `build-charts-from-signals.rb` runs offline |
 | `chart-specs.json` | PINNED `build-charts-from-signals.rb` output — captures all three fixes in one deterministic artifact (see assertions) |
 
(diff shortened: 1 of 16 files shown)


twells89:
### Follow-up: cosmetic fix + real-world coverage sweep

**Cosmetic** (`5a14fc7`): pivot value columns that were a Tableau calc surfaced by their internal id (`Calculation_summary`) now resolve to the calc's caption (`Summary $`).

**Coverage sweep** (`8a747fa`): pulled a diverse set of real public Tableau workbooks off GitHub (a state-symbol map, a traffic-analysis workbook, a 12-sheet sales dashboard) and ran the gap scanner + parser over each. Findings:
- Filter resolution held (5/5 on the sales dashboard), maps detected (3/3 region), no unhandled features.
- **Caught a real over-fire in the crosstab fix:** an Automatic-mark sheet with Measure Names on one shelf but the measure *values* on the other (`[Multiple Values]` pill) is a multi-measure **bar**, and my first pass turned it into a pivot-table (hit on a public "Barchart of accident factors" workbook). Fixed by detecting the Measure-Values pill and excluding it from the Automatic relaxation — carefully, at the shelf level, so a shelf holding the pill alongside a real measure still counts its measure (an earlier per-field attempt flipped time-series line charts to bar; reverted).
- Re-verified no line/bar regression vs the `main` baseline; EDNA still detects 44 crosstabs. Corpus fixture gains a `Bookings by Quarter Bar` over-fire guard. `corpus --check`: **4/4 pass**.

twells89:
### Additional live renders on CSA.TJ (confidence-hardening)

Built three more workbook shapes directly off `CSA.TJ.ORDER_FACT`, POSTed + rendered, then deleted — to exercise the new functionality beyond the first E2E:

1. **Multi-dimensional crosstab** (Order Channel › Ship Method nested rows × Order Status columns × Net Revenue + Quantity values, Automatic mark). Renders with correct **2-level nested row grouping, per-channel subtotals, and a Grand-total row + column** — the richer EDNA shape, not just the 1-dim case.
2. **Quick filter actually applied** — selecting `Order Status = Complete` on the rendered workbook collapsed the crosstab to just that column group (Grand total 119,038 → 114,399), proving a resolved quick-filter control drives the grouped table live (the first E2E only showed the controls present).
3. **Over-fire guard, live** — an Automatic-mark multi-measure sheet with the `[Multiple Values]` pill renders as an actual **bar chart**, confirming the measure-values discriminator keeps bars as bars (not pivots) end-to-end, not just at parse time.

No line/bar regressions; all four original fixes hold. I'm comfortable with the merge.

PR #207, opened 2026-06-28
DDMX epic build-layer: aggregate-dimension build (y9rd.13), aggregate-metric percent-format (y9rd.5), predicted-parity audit mode (y9rd.6)

twells89:
Two build-layer pieces of the DDMX epic (beads-sigma-y9rd), both in `build-charts-from-signals.rb`.

## 1. y9rd.13 — consume the `aggregate-dimension` workbookPattern (calc-on-calc bar)

A Tableau dimension that buckets an aggregate metric (`High Margin Flag = IF [Margin Pct] > 0.3 ...`, `Margin Pct = SUM(GP)/SUM(GR)`) can't be a Sigma grouping dim or a row-level DM column, so the converter (mcp #75) reports it as an `aggregate-dimension` pattern. The build layer had no consumer → the bar was viz-pruned.

- `load_aggregate_dims` / `agg_dim_for` / `build_aggregate_dim_helper` emit a hidden helper **chain**: INNER grouped at base grain (aggregate metric(s) + measure as group calcs); PASSTHRU sources INNER **with `groupingId`** and carries the bucket column; the chart sources PASSTHRU and re-aggregates per bucket.
- `build-workbook-spec.rb`: place chart-specs `data_elements` (hidden helpers) on the Data page — previously only `master` was emitted, so any helper-sourced chart 400'd `Dependency not found`.

**Why the chain:** a CHART source **silently drops `groupingId`** (verified via readback), so a chart sourcing the grouped inner reads base rows with the aggregate repeated → `Sum` fans ×row-count. Live on CSA.TJ: chart-direct = **$82,172,896** (×705); chain = **$116,557.3** (de-fanned, matches the Running Revenue Line endpoint).

## 2. y9rd.5 build half — percent-format attainment/share ratios on the chart measure path

An aggregate-ratio metric (Revenue Attainment = Sum(Net Revenue)/Sum(Revenue Goal)) plotted as a non-pivot tile measure was placed correctly (yAxis, verbatim formula) but formatted as **currency** ("$0.93") because the converter guesses $-currency. Extends the pivot path's existing goal/target percent-format rule (PR #203) to the chart measure path, firing last to override the currency guess; gated on an actual division. Live-verified: testbed `Attainment by Region` bar → 4 region bars at ~93% on a 0–100% axis.

This also confirms aggregate metrics plot as the tile **measure** (not a row column) in the non-pivot case; the converter collapse-completeness half of y9rd.5 was already verified done on `main`.

## Tests / gates
- New `test-aggregate-dimension.rb` (offline, 20 assertions) wired into `corpus-check.yml`.
- Converter regression `npm run regression -- tableau` → **10/10** (converter untouched).
- `lint-skills` + `check-shared` clean; pre-commit hooks pass.
- Full `synth-twb-e2e.rb` on real CSA.TJ data for both tiles; test DM+workbooks deleted after verify.

## Out of scope
Pre-existing control-lint drift (`ctl-param-metric-picker` vs `ctl-metric-picker-crosstabs`) — reproduces identically on `origin/main`. Tracked as beads-sigma-jwsf.


twells89:
### Added: y9rd.6 — pre-migration predicted-parity % + A–D band (tableau-assessment)

Third epic child, now in this PR (commit 63dc931). Audit-mode parity **prediction** from static `.twb` signal, before any conversion:

- `aggregate-complexity.rb` rolls the gap-scanner's status tiers up by feature **occurrence count** into a 0–100% predicted parity + A–D band (`auto`=1.0, `hint`=0.85, `manual`=0.5, `unhandled`=0.0; no complex features ⇒ 100%/A). Occurrence-weighting is the point: 50 auto + 1 unhandled → 98% (A); 1 auto + 1 unhandled → 50% (C).
- Explicitly distinct from the post-migration **measured** value-parity score (`verify-parity.rb`, y9rd.2).
- Surfaced in `complexity.json`, `shortlist.json` (+ printed table), and `readout.md` (shortlist + complexity tables).
- Reuses the converter coverage-matrix ground truth via the existing gap-scanner tiers — no new function list to drift.
- `test-predicted-parity.rb` (offline, 9 assertions) wired into `corpus-check.yml`.

This PR now closes the DDMX epic's remaining build-layer/assessment children: **y9rd.13**, **y9rd.5 build half**, **y9rd.6**. (y9rd.4 was separately resolved as already-satisfied; y9rd.14 is a P3 stretch left open.)

PR #255, opened 2026-07-02
tableau: Phase-1 B3 parser — KPI scorecard (Shape/Circle BAN) detection

twells89:
**Stacked on #254** (base = `tableau-phase1-b4-text-emit`). Merge #252 → #254 → this in order.

## What

The design doc assumed KPI detection "just works" for the benchmark — it doesn't. Every KPI tile in *Job Loss from Mass Deportations* (the 4 region "Total Job Losses" cards + the rail BAN) uses `<mark class='Shape'/>`, which the parser mapped to `scatter`, so `is_kpi` was **false** and the KPI path never fired. This is the prerequisite for the B3 composite emit.

## Change (`parse-twb-layout.rb`)

- New `parse_customized_label(ws)` — reads a worksheet's pane `<customized-label>` BAN (the "big number" on a Shape/Circle mark), returns `{label, value_font_size, annotation_runs}` when a large-font run (`>= KPI_BAN_MIN_FONT=16`) exists, else `nil`. Æ sentinel stripped; dynamic `<[…]>` runs flagged `ref=true`; invisible tiny-font filler dropped from the label.
- KPI detection: a Shape/Circle mark **with a BAN** now counts as `kpi_capable` — **narrow by design**, so an ordinary symbol scatter (no BAN) keeps flowing to scatter. The `is_kpi` gate (zero dims + ≥1 measure) is otherwise unchanged.
- `chart_kind_for`: `shape`/`circle` → `'kpi'` when `is_kpi` (else `'scatter'`).
- Surfaces `kpi_label` / `kpi_value_font_size` / `kpi_annotation_runs` on chart zones for the emit PR.

Detection only; the composite emit (label + big value + annotation, reusing B4's `text_body_from_runs`) is the next PR — same parser→emit split as #252→#254.

## Test

`test-kpi-scorecard-detection.rb` runs the **real parser** on a synthetic `.twb` with a Shape BAN scorecard (→`kpi`, label/fontSize/annotation extracted) **and** a real Shape scatter (→stays `scatter`, no false-positive). Verified on the real benchmark: the 4 region cards now detect as `chart_kind=kpi` (fontSize 26, label "Total Job Losses"). Regression suite green.

Bead: `beads-sigma-ubr5.7` (KPI-detection half; composite-emit half next)

### DIFF
--- a/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/parse-twb-layout.rb
+++ b/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/parse-twb-layout.rb
@@ -350,12 +350,59 @@ xml.elements.each('/workbook/datasources/datasource') do |ds|
   end
 end
 
+# ── Phase-1 B3: KPI scorecard <customized-label> (BAN) extraction ────────────
+# A Tableau "big number" scorecard often uses a Shape/Circle mark with the value
+# rendered as a <customized-label> BAN inside the pane, e.g.:
+#   <customized-label><formatted-text>
+#     <run fontcolor='#333' fontsize='10'>Total Job Losses</run>   ← label
+#     <run fontsize='26'><[…measure…]></run>                        ← BAN value
+#     <run fontcolor='#666' fontsize='8'>of U.S. total</run>        ← annotation
+#   </formatted-text></customized-label>
+# Returns {label, value_font_size, annotation_runs} when a BAN (a run whose font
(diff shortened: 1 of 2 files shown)


twells89:
Superseded by #269, which cherry-picks this B3-parser commit (plus B4 emit + B3 emit) cleanly onto current main — the original stack tangled after #254 squash-merged into a stale base branch. Closing in favor of #269.

PR #266, opened 2026-07-02
tableau-to-sigma: B1/B2/C2/D1 composition stage (repeated per-category cards)

twells89:
## What & why

Adds a **composition/style layer** to `tableau-to-sigma` — the "design" half the skill was missing. Numbers, chart-kind, and parity were solid; **extracting + emitting** tints, palette, styled text, and composite per-category cards was not, so design-heavy dashboards came out as a flat chart list.

This stage detects a **container repeated per category** (e.g. the four region columns) and emits N tinted `GridContainer` cards — each stacking a transparent hero KPI + %-of-total + annotations + a Top/Bottom-N bar + a threshold strip — plus a left rail and a wired control row. Validated two ways: reproducing a design-heavy hand-migrated dashboard, and a **cold run on a second, unrelated Tableau workbook** (real Superstore `.twb` + data landed fresh) that fell out end-to-end.

## Scope

- Plugin(s) touched: **tableau-to-sigma** only
- [x] Exactly one plugin. **Additive** — a new `scripts/composition/` dir; touches no existing or gated script.

New files under `scripts/composition/`:
- `detect-composition-signal.py` — decides applicability + **infers roles** (category via `.twb` color palette + cardinality; entity via cardinality; **measure via `.twb` reference-frequency**, not magnitude; rate; secondary; split; threshold via a `> N` flagging ~10-40%).
- `detect-region-cards.py` — EXTRACT: parse signals → `config.json` (field role map, palette from `.twb` color buckets, tints derived from the base hue, facts/cutoffs). `--twb` optional (data-only). Data-model **or** warehouse-table source.
- `compose-region-cards.py` — EMIT: config → composed workbook spec + embedded grid layout. Fully config-driven; N cards; optional split/secondary/threshold.
- `compose-auto.py` — **one-command auto-trigger** (signal → detect → emit; exit `2` = does-not-apply → route to standard flow).
- `composition-verify.py` — Phase-6 gate: structural completeness + **live per-card KPI vs config facts** (catches render races / broken tiles).
- `test-compose.py` + `config.example{,2,3}.json` — three schemas (region/worker-split/threshold · cloud no-split · warehouse-source no-threshold).

## Verified-live spec recipes (see `composition/README.md`)

Control-driven 3-way measure `Switch` · Most/Least-impacted via a **window-free** control-conditional cutoff boolean (`Rank()` errors as a window fn) · Show/Hide median via a `refMarks` conditional formula · C2 threshold color · container tints + `categoricalScheme` + white canvas. **Known spec gaps surfaced, not shipped as dead UI:** `dataLabel` toggle is a static enum; click-to-filter needs an added control + hidden base table.

## Evidence

- **Reproduction:** the design-heavy source dashboard rebuilt from the skill (region-scoped KPIs verified by query, controls wired).
- **Cold run (generality):** real second `.twb` (Superstore) — signal inference picked `measure=Sales` (352 `.twb` refs, high confidence), landed data cold, `compose-auto` → new workbook → **live-verify PASS**. Caught + fixed a real bug in the process (empty-body text elements render as error tiles → now dropped).

## Checklist

- [x] `ruby tools/check-shared.rb` — green (pre-commit hook: 265 shared-file copies match)
- [x] `ruby tools/lint-skills.rb` — green (pre-commit hook: 10 converter SKILL.md gates documented)
- [x] Phase numbers unchanged (documented as standalone Phase-5c entrypoint; not inlined into the gated `migrate-tableau.rb`)
- [x] No unrelated working-tree changes
- [x] `composition/test-compose.py` PASS (3 schemas) · tableau ruby offline suite 32/0

## Follow-ups (non-blocking)

- Auto-wire `compose-auto` into `migrate-tableau.rb` discovery (currently a documented standalone step).
- Palette extraction fires only when the source `.twb` encodes a categorical color map (defaults to brand-neutral otherwise).
- Secondary rail KPI carries a generic "Total" label.

### DIFF
--- a/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/composition/README.md
+++ b/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/composition/README.md
@@ -1,5 +1,25 @@
 # Composition stage (B1/B2/C2/D1) — repeated per-category cards
 
+## Applicability (read first)
+
+This stage reproduces **one specific source design**: a Tableau dashboard that lays
+out a **container repeated per category member** (the four region cards in the
+Job-Loss benchmark — `South Container`, `West Container`, … in the `.twb`, each with
+the same children). It is **not** a general design reconstructor. `compose-auto`
+only fires when `detect-composition-signal` finds that repeated container **in the
+source `.twb` zone tree** — a dashboard whose data merely has a category+entity+measure
+(e.g. Superstore) correctly routes to the standard build instead of getting this
+layout imposed on it. Sigma's spec has **no repeated-container primitive**, so the
+repeat is detected source-side and the emitter materializes N *explicit* GridContainers.
+
+**Current honest limit:** when it applies, the card's *children* (KPI + %-annotation +
+Top-N bar + strip) and the rail are still modeled on the Job-Loss source structure
+rather than derived per-dashboard from the matched container's own children. Deriving
+those from the source is the open follow-up; until then, treat this as faithful for
+the Job-Loss-class card-trellis and verify any new card design by eye.
+
+
+
 The data-correctness layer (right numbers, right chart kind, parity gate) doesn't
 reconstruct a design-heavy dashboard's *composition*. This stage does: it detects a
 **container repeated per category member** (the four region columns in the Job-Loss
--- a/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/composition/README.md
+++ b/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/composition/README.md
@@ -1,0 +1,89 @@
+# Composition stage (B1/B2/C2/D1) — repeated per-category cards
+
+The data-correctness layer (right numbers, right chart kind, parity gate) doesn't
+reconstruct a design-heavy dashboard's *composition*. This stage does: it detects a
+**container repeated per category member** (the four region columns in the Job-Loss
+benchmark) and emits N tinted `GridContainer` cards, each stacking a transparent hero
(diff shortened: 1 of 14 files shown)


twells89:
Converting to draft: the composition layout is a fixed template (the Job-Loss card-trellis) reskinned with any dataset's data, NOT a reconstruction of the source dashboard's design. The applicability gate fires on data shape (category+entity+measure) rather than on the source .twb actually being a repeated-per-category card design — so it mis-fires on non-card dashboards (e.g. Superstore). Reworking: (1) real repeated-container detection from the .twb zone tree as the gate, (2) derive card/rail structure from the source rather than the hardcoded skeleton. Do not merge as-is.

twells89:
Closing: this is a Job-Loss-specific card-trellis *template*, not the general design layer we actually want. The real goal is to carry each source .twb's own design signals (palette/tints/canvas/styled-text/structure) into the normal build so every migration looks like its own source — not to stamp one layout on every dashboard (which is exactly what this did: the Superstore run came out identical to Job-Loss).

Nothing of value is lost: the verified spec recipes (control-driven measure Switch, window-free rank cutoff, refMarks median toggle, C2 threshold color, empty-text-tile fix, warehouse-table source, measure-inference-by-.twb-reference-frequency) and the source-side applicability gate are all preserved in the session handoff (~/tableau-composition-handoff.md) and memory. The Job-Loss live reference workbook stays. Next: audit what the normal pipeline already emits (B2 tints #246 / D1 palette #248 merged; B4 text in-stack), measure the real design gap on Job-Loss + Superstore, then build a general extract→apply layer to that gap. Branch left intact (not deleted) in case the recipes are cherry-picked into a docs PR.

Issue #259, opened 2026-07-02
Migration fidelity: enforce visual/value gates + emit-side fixes (sequence after in-flight PRs)


twells89:
Follow-up to #258 (which added the Operating Contract docs). These are the **code** changes that make the contract enforceable. Split out because several touch **hot / shared** files (`shared/scripts/`, `build-charts-from-signals.rb`, `phase6-parity.rb`, `migrate-tableau.rb`) that in-flight converter PRs are actively editing — merge deliberately to avoid collisions.

Root cause across all of these: fidelity checks that exist but are **soft / opt-in**, plus a few emit-side gaps. Making them **hard + MCP-independent** converts "silently ships wrong" into "stops and tells you."

### 1. Visual-comparison gate → enforced by default  *(shared lib — highest collision)*
`shared/scripts/assert-phase6-ran.rb` gate **8b**: flip from opt-in (`--require-visual-comparison`) to **default hard-fail** on a missing source-vs-target visual verdict; escape hatch `--skip-visual-comparison "<reason>"`; keep `--require-visual-comparison` as a no-op for back-compat. Edit the **canonical** copy then `ruby tools/sync-shared.rb` (propagates to all 9 converters — intended: fidelity-by-default everywhere). Add a fill/coverage check: fail if any page's placed-tiles < zones or grid-fill < ~45%.

### 2. Layout fallback + dead safety net
`build-dashboard-layout.rb`: the bottom-band safety net's `kind` check (`%w[chart control]`) never matches real kinds (`bar-chart`/`kpi-chart`/`table`), so unmatched tiles vanish silently — fix the check; emit a `layout-census.json` (zones/placed/dropped/fill%) for gate #1. `migrate-tableau.rb` Phase 5: the three `WARN: … default stacked` branches should hard-fail for multi-tile dashboards.

### 3. Value fidelity (KPIs come out wrong)
KPI emission (`mechanical-specs.rb` / `build-charts-from-signals.rb`): prefer the source's **materialized validated calc columns** (`*(copy)*`) over re-deriving `SUM(x)*param`; translate the calc **kind** (parameter→control ref, ratio→division, LOD→grouping) from `extract-calc-fields.rb`; use the source's exact aggregate (`Count` vs `CountDistinct`) and population. Make `verify-parity.rb` a **value gate**: compare the number to the source view-data and fail on drift.

### 4. Controls / params dropped
[...]

PR #267, opened 2026-07-02
fix(tableau): KPI value fidelity — prefer validated materialized calc + resolve internal-name captions

twells89:
Part of #259 (TASK 1 — value fidelity: KPIs come out wrong). First increment.

## Problem

Tableau scorecard ("big number") tiles migrated with **wrong KPI numbers** — a value that should read thousands read millions. Root cause is two coupled emit-side bugs in `build_kpi_element` / `resolve_shelf_field`:

1. A scorecard's Marks card carries **several** measures: a raw aggregate column, one or more opaque `Calculation_NNN` calc ids, and the **materialized validated calc** the author actually trusts (named `[<Field> (copy)_NNN]`, caption ends `"(validated)"`). The old code took `measures.first` — usually the raw column — and emitted `Sum(rawcol)`, ignoring the validated per-row logic.
2. Those calc columns are named `[Calculation_NNN]` / `[<Field> (copy)_NNN]`, **not** 36-char GUIDs, so `guid_from_text()` returned `nil` and `resolve_shelf_field`'s guid lookup missed. The field then fell back to the raw internal name, so both `map_column` **and** the calc-formula ladder missed → forced the naive `Sum(rawcol)` even when the caption (and its validated formula) was available in `columns_by_guid`.

## Fix

1. **`pick_kpi_measure`** (new pure helper): prefers the validated/materialized calc (`(validated)` caption > `(copy)_NNN` name > any `User`-derived calc > raw aggregate), never picks a `(Label)` text calc as the value, order-stable tie-break. Wired into `build_kpi_element` in place of `measures.first`.
2. **`resolve_shelf_field`**: added a bracket-stripped **internal-name fallback** into `columns_by_guid`, so a `(copy)_NNN` / `Calculation_NNN` column resolves to its real caption (the `"(validated)"` calc). The existing formula-resolution ladder (ratio → division, parameter → control ref, LOD → grouping, user-agg → decompose) then fires as intended.

Together these route a KPI tile to its validated calc's real formula instead of a naive sum of a raw column.

## Scope / follow-ons (tracked in #259)

- This lands the **selection + caption-resolution** foundation. The per-class formula translation it unblocks (ratio-of-two-materialized-cols → `Sum(a)/Sum(b)`; parameter-scalar → control ref) rides on the existing ladder; any remaining gaps there are a follow-on.
- The **value-parity hard gate** (fail on drift vs source view-data) and the **rename-materialized-calc-to-caption** item (queue #2) are separate PRs.

## Test

`test-kpi-value-fidelity.rb` — 9 checks on genericized fixtures mirroring the real marks-card shape (raw col + opaque calc + `(copy)_NNN` validated + `(Label)` copy). Full tableau suite green (29/30; `test-calc-discovery` pre-fails on `main` — needs `TABLEAU_AUTH_TOKEN`, unrelated). `tableau-only` files — no shared re-vendor.

### DIFF
--- a/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/build-charts-from-signals.rb
+++ b/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/build-charts-from-signals.rb
@@ -1773,11 +1773,28 @@ SHELF_TRUNC_FOR_PREFIX = {
 }.freeze
 
 def resolve_shelf_field(field, meta, mmap)
+  cols_by_guid = meta['columns_by_guid'] || {}
   guid = field['guid']
   cap_for_field = nil
   if guid
-    info = (meta['columns_by_guid'] || {})[guid]
+    info = cols_by_guid[guid]
     cap_for_field = info && info['caption']
+  end
+  # Bracket-stripped internal-name fallback (bead: KPI value fidelity). Tableau
+  # calc columns are named `[Calculation_NNN]` or `[<Field> (copy)_NNN]` — NOT a
+  # 36-char GUID — so guid_from_text() returns nil and the guid lookup above
+  # misses. But `columns_by_guid` IS keyed by that internal name and carries the
+  # real caption (e.g. a "…(validated)" calc). Without this the field falls back
+  # to the raw `(copy)_NNN` string, map_column + the calc-formula lookup both
+  # miss, and the KPI naively re-derives `Sum(rawcol)`.
+  if cap_for_field.nil?
+    raw_key = field['raw'].to_s
+                          .sub(/^\[[^\]]+\]\./, '')
+                          .gsub(/^\[|\]$/, '')
+                          .sub(/^[a-z]+:/i, '')
+                          .sub(/:[a-z]+$/i, '')
+    info2 = cols_by_guid[raw_key]
+    cap_for_field = info2 && info2['caption']
   end
   cap_for_field ||= field['raw'].to_s
                                  .sub(/^\[[^\]]+\]\./, '')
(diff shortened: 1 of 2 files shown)


twells89:
### Validation against the live reference build (read-only)

Cross-checked this change against the hand-finished reference workbook's own spec (read-only `GET /spec`) and its live values (read-only element export). Both corroborate the fix's direction:

**Answer key (the expert's own KPI formulas):** every scorecard KPI sums the **materialized validated `(copy)` calc column**, composed via ratio / difference / count / parameter — never a raw `Sum(rawcol)`. Pure-Sum KPIs are exactly `Sum([… <Field> (copy)_NNN])`, which is precisely what `pick_kpi_measure` now selects (verified matching for those KPIs).

**Live ground-truth** (async element export → poll → download) confirms the target numbers the composed formulas produce (ROI ≈ 4.66×, avg-cost ≈ $378.8, a param-scalar KPI = its parameter value, etc.). This export path (`POST /v2/workbooks/{id}/export` → `GET /v2/query/{qid}/download`) is a good ground-truth source for the value-parity gate when source CSVs aren't on disk.

**Confirmed follow-on (not in this PR):** the **ratio/difference** KPIs still fall to the naive fallback. `translate_user_agg_formula` only rewrites *explicit* `SUM([x])`; a validated calc written as `[measA (copy)]/[measB (copy)]` is **bare** measure refs, so the regex misses, the residue check sees unresolved `[...]`, and it returns nil. Fix = a KPI-scoped step that wraps bare materialized-measure refs in `Sum()` and maps them to the DM column (`Sum([Master/A])/Sum([Master/B])`, the answer-key form). Deferred because it needs a live-pipeline E2E to verify the DM actually carries the `(copy)` columns (ties to the materialize-and-rename queue item). This PR is the selection + caption-resolution foundation that unblocks it.

twells89:
### Ratio/composite fix added + live-verified

Pushed a follow-on commit on this branch: `translate_kpi_measure_formula` + internal-name ws_calc matching, so ratio/difference KPIs (ROI, avg-cost, net-savings) emit `Sum([Master/A]) / Sum([Master/B])` instead of dropping to a naive `Sum(rawcol)`.

**Verified end-to-end against the real warehouse** (read-only): built a throwaway scratch workbook via `POST /v2/workbooks/spec` sourcing the same data model, wrote the *exact* forms this translator emits, and exported them. They query to the source's own values:
- ratio KPI → **4.663365**
- avg-cost KPI (my `Sum(…)/CountIf(IsNotNull(id))` form) → **378.845487564124** (== the hand-build's `Count(id)` form)
- pure-Sum KPI → **6,323,000**

Scratch workbook deleted afterward; the source workbook was never modified (confirmed 200 after cleanup).

**One dependency for full E2E:** the master element must actually **carry** the materialized `(copy)` columns — that's the model-side (mechanical-specs materialization + the rename-to-caption queue item), tracked separately. This PR is the complete emit-side half, now backed by a live number match.

Test now 14 checks. Full tableau suite green (29/30; `test-calc-discovery` pre-fails on `main`).

PR #268, opened 2026-07-03
tableau: Phase-1 B4 emit — styled static-text elements + layout placement

twells89:
Brings the B4 styled-text **emit** onto main. (Originally #254; it squash-merged into its stale base branch instead of main due to a base retarget miss — this re-lands it cleanly against current main.)

Emits styled static-text elements (color/bold/size/align runs) + their layout placement in `build-charts-from-signals.rb` / `build-dashboard-layout.rb`, with offline tests.

Part of the general Tableau→Sigma design-signal layer.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


twells89:
Superseded by #269. This branch's merge-base predated the squashed #252, so it re-applied the parser and conflicted. #269 is the clean integration. Closing.

PR #271, opened 2026-07-03
tableau design-signal APPLY: source palette + source-derived header chrome

twells89:
## What
Generalizes the workbook theme derivation so **each migration comes out looking like its own source** at the palette/canvas layer.

Surfaced by the multi-dashboard benchmark: `derive_theme` built `categoricalScheme` **only** from 8-digit-alpha container fills — the Job-Loss region-card *tint* idiom. Any dashboard whose design is a **color scheme** rather than tinted cards (ER's monochromatic reds, Udemy's subject-dot palette) has no such containers, so the theme degenerated to picking **white card backgrounds** as the palette and white as the canvas.

## Change
- **parse-twb-layout.rb**: emit `brand_palette` per dashboard from the color **encoding** (`<map to='#..'>`) + custom `<color-palette>`, frequency-ranked, neutrals dropped. Neutral = low chroma OR near-white (all channels high) OR near-black — tested on `mn`/`mx` so a vivid single-near-max-channel colour (orange `#f06719`, red `#ff0000`) is **kept**.
- **build-workbook-spec.rb** `derive_theme`: prefer `brand_palette` for `categoricalScheme` (falls back to the container-tint palette when the source encodes no brand colours); strip 8-digit alpha off the canvas → solid `#rrggbb`.

## Validated (4-dashboard benchmark)
| Dashboard | Derived palette |
|---|---|
| ER Visits | `#ba2020` + `#8f1716` + red ramp (monochromatic red) |
| Quota | `#0088ff` blue primary + teal |
| Tableau History | `#4e79a7` blue + `#b9ddf1` (Water) + `#f06719` (Gold) |
| Udemy | `#4e79a7`/`#e04053`/`#54a7ae` subject-dot colours |

Each source yields its **own** palette instead of white.

## Tests
- new `test-brand-palette.rb` (extraction, freq-rank, dedup, + the near-white neutral bug)
- extended `test-theme-derivation.rb` (brand_palette beats white fills; alpha-strip; single-colour falls back to tints)
- both gated in `corpus-check.yml`; full offline suite green

Part of the general Tableau→Sigma design-signal layer.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/corpus-check.yml
+++ b/.github/workflows/corpus-check.yml
@@ -107,6 +107,8 @@ jobs:
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-styled-text-layout.rb
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-kpi-scorecard-detection.rb
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-kpi-composite-emit.rb
+            plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-theme-derivation.rb
+            plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-brand-palette.rb
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-element-id-namespacing.rb
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-fact-pick.rb
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-layout-lint.rb
--- a/.github/workflows/corpus-check.yml
+++ b/.github/workflows/corpus-check.yml
@@ -109,6 +109,7 @@ jobs:
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-kpi-composite-emit.rb
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-theme-derivation.rb
             plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-brand-palette.rb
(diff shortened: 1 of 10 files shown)


twells89:
Added a second improvement to the general design-signal APPLY layer, surfaced by the multi-dashboard benchmark render:

**Source-derived header chrome** — the layout builder used to stamp a fixed dark-navy header band (`HEADER_STYLE`) with forced-white title text on *every* dashboard, so a minimalist white source (Tableau History) came out looking nothing like itself. Now `build-dashboard-layout` derives the header from the source zone tree: a colored band is emitted **only** when the source has a band-like fill (dark or saturated) in the header region; otherwise the title sits on the page canvas (transparent). Verified live: History now renders as a dark title on white (matches source) instead of a navy bar. `test-header-derivation.rb` added + CI-gated.

Both commits are the same theme — carry the source's own design into the output — verified end-to-end on a live Sigma workbook.

twells89:
⚠️ **Not mergeable as-is — breaks `test-container-tint.rb`.** On this branch head (2285e30) the test fails: `page header container still emitted (unaffected)` (passes on main). Root cause: this PR replaces the fixed dark `HEADER_STYLE` band with `header_from_source()` (transparent when the source has no header band) — a deliberate, good behavior change, but the affected test wasn't updated to match. CI showed green at an earlier commit; the current head regressed it. Fix before merge: update `test-container-tint.rb`'s header assertion to the new source-derived behavior (or give its fixture a header-band fill), then re-run the full offline suite. Held out of the 2026-07-05 merge batch (#272/#267/#270 merged; #273 opened) for this reason.

PR #272, opened 2026-07-03
fix(controls): don't fail the control gate on already-surfaced coverage gaps

twells89:
Part of #259 (controls track). Found by running a filtered dashboard (RLS Reference) end-to-end.

## Bug — a stray parameter blocks the whole control gate

`control_lint`'s "missing control" check hard-failed whenever a `control-scope.json` entry had no matching control in the spec — **including entries the builder intentionally records as gaps rather than placing:**

- **`needs-wiring`** — an orphan/unreferenced Tableau parameter (e.g. a stray `Year Filter` that no calc references). The builder emits it *for coverage*, and the page-assembly dead-control guard correctly **skips placing** it (a control that changes nothing shouldn't ship as furniture) — but the scope record survived, so the gate failed with *"missing control … the source filter was not migrated."*
- **`needs-materialization`** — a calc-bound filter whose column isn't on the model yet (surfaced for the operator to materialize).

Both are recorded in the controls-coverage ledger — **not silent drops** — so their absence from the spec is expected, not a failure. As-is, **any dashboard carrying a stray parameter** (extremely common) died at the control gate. The `jwsf` fix had covered only *measure-picker* params (which are referenced, so they get placed); plain orphan value params still tripped it.

## Fix

In the scope↔spec reconciliation, skip entries whose `status` is `needs-wiring` / `needs-materialization`. Only an **`emitted`**-status entry missing from the spec remains a real "not migrated" failure (regression-guarded).

## Shared — fixes every converter

Edited canonical `shared/lib/control_lint.rb` + `tools/sync-shared.rb` fan-out to all 7 converter copies, so **every** converter's control gate is fixed (not just Tableau) — relevant for the hackathon. New shared regression test `scripts/test-control-lint.rb` (4 checks: needs-wiring/needs-materialization suppressed; emitted + default-status still flagged), fanned out to powerbi + tableau.

## Verification

- `check-shared` OK (267 copies consistent); control/param test suite green (`test-control-lint`, `test-param-measure-picker`, `test-coverage-gate`, `test-calc-field-controls`, `test-control-display-type`).
- Against the **real posted RLS-Reference workbook** (built E2E into a sandbox org): `control lint` now **clean (3 controls checked)** — the orphan-param FAIL is gone, so the migration passes the control gate.

### DIFF
--- a/plugins/looker-to-sigma/skills/looker-to-sigma/scripts/lib/control_lint.rb
+++ b/plugins/looker-to-sigma/skills/looker-to-sigma/scripts/lib/control_lint.rb
@@ -270,7 +270,16 @@ module ControlLint
     end
 
     # (c) annotated controls that never made it into the spec ------------------
+    # A scope entry flagged as an ALREADY-SURFACED gap (needs-wiring: an orphan/
+    # unreferenced parameter the builder intentionally does not place as a dead
+    # control; needs-materialization: a calc-bound filter whose column isn't on
+    # the model yet) is recorded in the controls-coverage ledger, NOT silently
+    # dropped — so its absence from the spec is expected, not a "not migrated"
+    # failure. Only a control the builder recorded as EMITTED but that's missing
+    # from the spec is a real bug the source filter was lost.
+    surfaced_gap = %w[needs-wiring needs-materialization].freeze
     scope_by_cid.each do |cid, c|
+      next if surfaced_gap.include?(c['status'].to_s)
       violations << "missing control: control-scope.json expects control #{cid.inspect}" \
                     "#{c['sourceName'] ? " (source: #{c['sourceName'].inspect})" : ''} but the spec " \
                     'has no control with that controlId — the source filter was not migrated'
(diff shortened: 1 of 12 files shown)


twells89:
### Hardening pass — 3 filter-diverse dashboards + a live flip-test

Ran a filter-shape matrix through the full pipeline (into a sandbox org) to validate control emission/wiring, not just the gate:

| Dashboard | Filter shapes exercised | Result |
|---|---|---|
| **RLS Reference** | categorical quick-filters, relative-date, RLS calcs, 1 orphan param | Controls emit + **structurally wire** (control_lint clean); RLS detected + flagged (not applied) |
| **Sales Ops (Multi)** | measure-picker param (Metric Switch), **105 orphan params**, categorical, relative-date | Metric Switch → `emitted`, mechanism=formula; 105 needs-wiring surfaced, **0 unaccounted**; control gate passes (validates this fix at scale) |
| **Complex Pivot** | pivot + orphan numeric-range param, window pivot values | **Surfaced the 2nd facet fixed in `c453cfb`**; now lints clean |

**Live flip-test (gold standard)** on the posted RLS workbook — the emitted Region control actually filters the data:
- `ctl-region = West` → chart returns only **West (17,535.46)**
- `ctl-region = South` → chart returns only **South (15,343.76)**

So: orphan params (the common real blocker) no longer false-fail the control gate in **either** rule, controls emit + wire + filter live, and gaps are surfaced honestly (0 unaccounted). The pass paid for itself by finding the "no controls built" facet a single dashboard wouldn't have.

PR #412, opened 2026-07-17
tableau: derive real db/schema from the .twb (stop defaulting to CSA.TJ) — #7a

twells89:
## Why (from TWO live sessions)

A live-warehouse workbook whose db.schema the operator didn't pin defaulted to the **CSA.TJ placeholder**, which **404'd on every Sigma catalog sync**. The real db.schema was only recoverable by the agent hand-reading the `.twb`, forcing a manual `--db/--schema` re-run. Backlog **#7a**.

## What (orchestrator-only, no shared files, no converter edit)

- **`MechanicalSpecs.derive_db_schema_from_twb(twb)`** — self-contained REXML parse returning `[db, schema]` from the first **live-warehouse** datasource's table relation (3-part path wins; else conn `dbname` + 2-part schema; else conn `dbname`/`schema`). **Never** derives for embedded-file extracts (CSA.TJ is the correct landing target there), basemaps, or sqlproxy stubs (hydrate resolves those). **Never fabricates** — returns `[nil,nil]` when unresolved. Self-contained so it doesn't `require` the un-main-guarded `scan-workbook-gaps.rb`.
- **`migrate-tableau.rb`** — before the extract/hydrate/converter consumers, when the user did NOT pass `--db/--schema` and the source isn't sqlproxy, set `opts[:db]/opts[:schema]` from the derivation. All four consumers inherit it; `--db/--schema` still win; embedded still lands at CSA.TJ.

## Test
`test-db-schema-propagation.rb`: 7 cases (2-/3-/1-part paths, embedded skip, sqlproxy skip, no-fabrication, missing file). Regressions green (multi-ds, extract-table-mapping, converter-fixtures); `check-shared` clean.

Corroborated on both reviewed DraftKings-class runs. Pairs with #411 (function leaks).

### DIFF
--- a/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/mechanical-specs.rb
+++ b/plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/mechanical-specs.rb
@@ -87,6 +87,59 @@ module MechanicalSpecs
     formula.to_s =~ /\[[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\]/i
   end
 
+  # #7a — derive the REAL [db, schema] from a .twb's first LIVE-warehouse
+  # datasource table relation, so the DM/converter/land steps stop defaulting to
+  # the CSA.TJ placeholder (which 404s on catalog sync — field-caught on 2 live
+  # runs; the real db.schema was only recoverable by hand-reading the .twb).
+  # Returns [db, schema] or [nil, nil]. NEVER derives for embedded-file extracts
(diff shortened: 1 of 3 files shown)


twells89:
Closing in favor of #414, which owns the db/schema line (manifest-derived db/schema + sqlproxy resolver) and is fixing the fallback-default regression directly. This PR's .twb-derivation also does not cover the case that matters most — a publishedConnection (Virtual Connection) workbook, where the real path lives inside the relation *name* (`ORDER_FACT (CSA.ORDER_FACT)`) rather than in `dbname`/`schema` attrs — so it's both redundant with and insufficient vs #414. Superseded.

PR #413, opened 2026-07-17
shared/preflight: flag unbounded number-range controls (C7) — [B] subset

twells89:
## Why (Opus 4.8 run)
The builder emitted `number-range` controls with `min:null/max:null` that passed offline lint but **400'd at POST** with an opaque union-type error mislabeled "expected modal" — a control-shape class that slipped past the C6/C7/C8 validator from #408.

## What
Extend **C7** (`shared/lib/preflight_lint.rb`): a `number-range` control with neither `min` nor `max` is an unbounded range that 400s — flag pre-POST with the fix. Mirrors the range-slider `low`/`high` rule. Synced canonical → 9 plugins.

## Test
`test-preflight-lint` +2 (unbounded → violation; bounded → clean). `check-shared` + all preflight tests green.

## Scope note (honest)
The other two classes need info not in the pre-POST spec, so they are deliberately **not guessed** here: list value-source from a calc/boolean column (needs DM column-type), control filter targeting a hidden/integer-coded master column (needs element-visibility + type; overlaps control_lint), and the "let parity proceed past unwired controls" waiver (a policy call). Flagged for a design decision.

### DIFF
--- a/plugins/cognos-to-sigma/skills/cognos-to-sigma/scripts/lib/preflight_lint.rb
+++ b/plugins/cognos-to-sigma/skills/cognos-to-sigma/scripts/lib/preflight_lint.rb
@@ -33,7 +33,8 @@
 #   C6  a control with an unknown controlType, or the docs-only `top-n` type that
 #       the live tenant 400s — mirrors Sigma spec/verify offline (canary 2026-07-11).
 #   C7  a control missing a REQUIRED field for its type: `mode` on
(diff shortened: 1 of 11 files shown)


twells89:
Closing to keep #414 the single line. #414 fixes the root cause (the emitter now parses <range> and no longer emits an unbounded number-range), which makes this preflight guard belt-and-suspenders. The one-rule C7 check (number-range with neither min nor max) can ride along in #414 if a defensive guard is still wanted.

PR #420, opened 2026-07-18
tableau: join-cardinality probe + join ledger — prove "right unique on keys" before Lookup ships (gate 16, exit 23)

twells89:
> Base is `integration/v5.5-e2e` (main + #414). Retarget to `main` after #414 squash-merges. Mandatory before the next field run (PLAN-v3 PR-4).

**Problem.** Sigma's `Lookup()` returns one arbitrary match per key. The v5.6 join-coalesce synthesis automates `Coalesce([X], Lookup([Target/X], key...))` across federated DM elements with no uniqueness check — in a field migration the target table was one grain finer than the join key, so every aggregate over the looked-up column silently undercounted. Mirror risk: agents delete "no-op" LEFT JOINs without proof (fan-out).

**Fix — measure, then gate.**
- **Ledger** (`scripts/lib/join_plan.rb`, wired into the orchestrator right after `dm-spec.json` is written): one `join-plan.json` entry per federated `.twb` join + per synthesized `Lookup()`, each `status: "unprobed"` with `grain_assumption: "right unique on keys"`. Empty ledger still written — presence is gate evidence. Vendored converter untouched (derivation lives in the Ruby layer per the portability ADR).
- **Probe** (`scripts/probe-join-keys.rb`): `GROUP BY keys HAVING COUNT(*)>1 LIMIT 5` + `COUNT(*) vs COUNT(DISTINCT keys)` through the established probe-workbook Custom SQL seam; `--fixture DIR` offline mode. Non-unique -> FATAL naming the entry, sample duplicate keys, and the two sanctioned resolutions (pre-aggregate the target to key grain + repoint the Lookup, or operator escalation), recorded in-ledger via `--resolve`.
- **Gate 16 / exit 23** (shared `assert-phase6-ran.rb`, synced): unprobed or unresolved non-unique -> fail; missing ledger + `Lookup(` in dm-spec -> fail. No skip flag — the recorded resolution IS the escape, so it never touches the waiver budget.

**Tests:** `test-join-plan.rb` (27), `test-probe-join-keys.rb` (34), +8 gate scenarios in `test-assert-phase6-gates.rb`; `test-join-coalesce-synthesis.rb` untouched and passing; check-shared / lint-skills / hygiene / corpus green. Fixture vocabulary neutralized per the private guard set.


twells89:
**Live e2e follow-up: two defects found and fixed** (commit 3ee9724)

**E2e evidence**

1. **VC workbooks produced unprobeable ledger entries.** Running the probe against a workbook backed by a Tableau published/virtual connection: the .twb carries no metadata-records, so federated-join entries derived `right_table` as the VC inode FQN (`<inode-guid>.NAME (X.TABLE)`) and `probe_keys` as Tableau GUID column ids. `probe-join-keys.rb` then errored on every entry (exit 3) instead of measuring cardinality — the gate blocked with no path to proof.
2. **`--reuse-dm` bypassed gate 16 entirely.** The ledger derivation lived inside the `unless reuse_dm_id` build block, so a reuse run never wrote `join-plan.json`; gate 16's belt-and-braces only greps `dm-spec.json` for `Lookup(` — a file the reuse path never writes — so a reused DM over a federated source went GREEN with zero grain proof.

**Fixes**

- `lib/join_plan.rb`
  - `right_table`: a VC relation label (`NAME (X.TABLE)` / `NAME (DB.SCHEMA.TABLE)`) now yields the parenthesized physical path, completed with the run's resolved `--db/--schema` (new `db:`/`schema:` args on `JoinPlan.derive`) into a probeable `db.schema.table` FQN.
  - `probe_keys`: GUID-like join keys resolve to physical columns via the .twb caption, folded with the same upcase+underscore folding the phantom-column filter uses (`mechanical-specs.rb fixup_dm_spec` convention — no divergent helper).
  - Custom-SQL Lookup targets record `right_sql` (statement); `right_table` stays null.
  - When neither resolution works, the old values are kept — the probe errors and the gate keeps blocking (safe direction).
- `probe-join-keys.rb`: entries with `right_sql` are probed as a subquery — `SELECT <keys>, COUNT(*) FROM (<right_sql>) t GROUP BY … HAVING COUNT(*) > 1 LIMIT 5` plus the totals pair.
- `migrate-tableau.rb`: the ledger derivation now runs on **both** paths. On reuse it scans the LIVE DM readback spec (`dm_spec_rb`, already fetched for id-mapping) for `Lookup(` synthesis plus the .twb for federated joins, landing in the same `<workdir>/join-plan.json`. Gate 16 (`assert-phase6-ran.rb`, shared canonical) is unchanged.

**Tests**

- `test-join-plan.rb`: VC label + GUID key resolution (new `test-fixtures/join-vc.twb`, invented names): 2-part paren path + run db, `DB.TABLE` + run schema, 3-part path as-is, safe fallback with no db/schema, plain federated .twb unaffected by the new args; sql-target entry records `right_sql` with `right_table` null.
- `test-probe-join-keys.rb`: `right_sql` fixture case — both probe SQLs wrap the statement as `FROM (…) t`, unique verdict, exit 0.
- `test-join-plan-reuse.rb` (new): wiring guards (one shared derivation call, reuse scans `dm_spec_rb`, sits outside the build-only block, writes the gate's exact path) + behavioral proof that a readback-shaped spec + .twb produce the ledger file with both entry kinds, status unprobed.

All touched tests pass (test-join-plan, test-probe-join-keys, test-join-plan-reuse, test-assert-phase6-gates, test-join-coalesce-synthesis, test-reuse-*, test-no-default-dbschema, + the 12 other migrate-tableau source-level guards); `tools/check-shared.rb` OK (374 copies in sync — no shared twin touched); hygiene sweep clean.

twells89:
Follow-up fix for a defect found by the Twin B live e2e run: with no `--folder-id`, `probe-join-keys.rb` POSTed `folderId: null` to `/v2/workbooks/spec`, which the API rejects with a bare 400 — the ledger recorded `status: error` with no hint at the real cause.

Fixed in 97e7725:
- the `folderId` key is now omitted entirely when `--folder-id` is absent (the API defaults to My Documents; same conditional stamp as `mechanical-specs.rb`)
- a failed probe POST now records the HTTP status **and** a response-body excerpt in `probe_error` and the console (the old `lines.first` truncation dropped the body), plus a console hint naming `--folder-id` when the error mentions the folder
- test coverage: stubbed-`sigma_rest` live-path cases asserting the no-folder-id POST body carries no `folderId` key, the `--folder-id` stamp, and the error surfacing (suite: 50/50 pass)

PR #511, opened 2026-07-28
feat(tableau-to-sigma): comparative KPI cards + prior-measure detection

twells89:
## What & why

Routes the Tableau builder's KPI emission through the shared `KpiCard.build` emitter (#510) and adds `lib/kpi_comparison_detect.rb`. When a KPI tile has a prior-period/target sibling measure, it wires a Sigma comparison delta — built with the **same `Sum([Master/..])` aggregation as the value column**, so the delta is `Sum(current) − Sum(prior)` by construction (verified live: 140 / 60 over multi-row data). Detection is conservative (exactly-one-match); the "comparison is UI-only" degradation warning is demoted only when a delta is actually wired. The shared emitter's native `name` object (vs a bare string) is propagated to `build-dashboard-layout.rb`'s four element-name matching sites via a new `el_display_name` helper. An integration test drives the real `.twb → parse-twb-layout.rb → build-charts-from-signals.rb` and asserts the emitted `comparisonColumn`.

**Stacked on #510** (shared `kpi_card` foundation) — merge that first.

## Scope

- Plugin(s) touched: **tableau-to-sigma** (only)
- [x] This PR touches exactly one plugin

## Checklist

- [x] `ruby tools/check-shared.rb` — green (515; no shared-lib change here)
- [x] `ruby tools/lint-skills.rb` — green
- [ ] `./corpus/run-corpus.sh --check` — **not yet run.** This changes a builder (`build_kpi_element`, `build-dashboard-layout.rb`), so KPI cases with a prior-period measure will now emit a `comparisonColumn`; goldens for affected cases likely need reconversion. Flagged for a reviewer/maintainer run before merge.
- [x] Phase numbers unchanged
- [x] No unrelated working-tree changes (fresh worktree off the shared branch)

Regression coverage: `test-kpi-comparison-wiring.rb` (detector units), `test-kpi-comparison-emit.rb` (real-pipeline integration), plus updated `test-kpi-param-scope.rb` / `test-signal-quality-floor.rb` / `test-display-title-match.rb` for the name-object change; existing `test-kpi-value-fidelity.rb` / `test-kpi-scorecard-detection.rb` green.

### DIFF
--- a/.github/workflows/corpus-check.yml
+++ b/.github/workflows/corpus-check.yml
@@ -292,6 +292,8 @@ jobs:
             plugins/quicksight-to-sigma/skills/quicksight-to-sigma/scripts/test-metric-reference.rb
             shared/lib/test_metric_binding.rb
             shared/lib/test_kpi_card.rb
+            plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-kpi-comparison-wiring.rb
+            plugins/tableau-to-sigma/skills/tableau-to-sigma/scripts/test-kpi-comparison-emit.rb
             plugins/domo-to-sigma/skills/domo-to-sigma/test/test-build-dm.rb
             plugins/domo-to-sigma/skills/domo-to-sigma/test/test-build-workbook.rb
             plugins/domo-to-sigma/skills/domo-to-sigma/test/test-convert-beast-modes.rb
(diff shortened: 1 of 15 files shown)


twells89:
**Corpus gate (offline `run-corpus.sh --check tableau`): 7/7 cases pass.**

Also confirmed none of the existing tableau corpus cases contain a KPI tile with a prior-period/target comparison measure (`grep -rilE 'comparisonColumn|vs.?prior|prior.*year' corpus/tableau/` → 0 hits). This change is behavior-preserving for single-measure KPIs and only emits a `comparisonColumn` when a prior sibling measure is detected, so it **does not stale any stored golden** — no `--reconvert` / golden update required. (A future live `--reconvert` is the way to add a dedicated KPI-comparison corpus case.)

PR #582, opened 2026-07-31
chore(powerbi): re-vendor converter — literal masking + lookup declutter (mcp#120, #121)

twells89:
Brings the Power BI bundle from `c3f892c` up to mcp `main` `2f8c6cd`. **Until this lands,
neither upstream change reaches any operator** — the skill runs the vendored bundle.

## What operators get

**mcp#120** — DAX formula scans no longer treat string-literal text as live syntax. A literal
containing a mapped function name or a bracketed name could be rewritten, and a reference
inside a literal could be read as a real column.

**mcp#121** — the reported issue: the data model arrived with hundreds of lookup columns to dim
columns *alongside* relationships to the same dims. Measured on four reference models:
**705 / 986 / 438 / 374** related columns against only 22 / 29 / 10 / 12 relationships.

They are now `hidden` and grouped **one folder per relationship target**, with names
disambiguated where a role-playing dimension produces several relationships to the same table.
**Nothing is dropped** — total column counts are unchanged, so every workbook binding still
resolves through the View alts. Narrowing was measured to be impossible converter-side (0 of
2,503 related columns are referenced by any converter-emitted formula); pruning the ~90% no
visual touches needs report-side signals and is a separate opt-in change.

## Verified

Imported **this branch's** bundle (not the source repo) and converted a reference model:
`940 total columns / 438 hidden / 10 folders` — byte-matching the source tree — plus 14
`TRIAGE:` warnings. 39/39 ruby suites across both `scripts/test-*.rb` and `tests/*.rb`.
Provenance and version-bump gates pass locally.

## A gotcha worth knowing before the next re-vendor

`tools/vendor-converters.sh` bundles from **`build/`**, not `src/`. A stale `build/` is vendored
**silently** while `PROVENANCE.json` is still stamped with the current git commit — the first
run here produced `hidden=0 / folders=0` while reporting `source 2f8c6cd`. So a current
`source_commit` does **not** prove a current bundle.

Run `npm run build` in the converter repo first, and verify behaviourally rather than by the
recorded commit. This also means a staleness check based only on comparing `source_commit`
against upstream would miss this failure entirely.

### DIFF
--- a/.github/workflows/hygiene.yml
+++ b/.github/workflows/hygiene.yml
@@ -165,6 +165,23 @@ jobs:
         run: |
           if [ -z "${RANGE_BASE:-}" ]; then echo "no diff range to check — skipping"; exit 0; fi
           bash tools/check-converter-provenance.sh "$RANGE_BASE" "$RANGE_HEAD"
+      - name: Converter-freshness guard (vendored-bundle staleness by age)
+        # Standing gate, NOT diff-scoped (unlike the pairing guard above, this
+        # runs every time regardless of RANGE_BASE): a merge to
+        # sigma-data-model-mcp changes nothing for any operator until someone
+        # runs tools/vendor-converters.sh and commits the regenerated bundle,
+        # and pre-existing drift nobody re-vendored in a long time has no diff
+        # to catch it on. CI cannot assume network access to that private repo,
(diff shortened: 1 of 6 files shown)


twells89:

PR #583, opened 2026-07-31
chore(qlik): re-vendor converter — literal masking + lookup declutter (mcp#120, #121)

twells89:
Bundle was pinned at `867c669`, so this picks up **more than today's fixes** — flagging that
rather than implying a one-change delta.

**mcp#120** stops Qlik expression scans from reading string-literal text as live syntax. Two
demonstrated defects: a literal containing a mapped function name had its content rewritten
(`'RangeSum(1) legacy code'` → `"(Coalesce(1, 0)) legacy code"`), and an unbalanced paren
inside a literal **silently dropped a real, separate translation** later in the same string.
Set Analysis clauses with bracketed or quoted field names now resolve back to raw text before
the field-name match — a field name is syntax, not data, and masking it made the whole measure
vanish.

**mcp#121** hides the derived View's related lookup columns and folds them one folder per
relationship target, dropping none.

Markers in the regenerated bundle: `maskQlikLiterals` 0→4, `qlikResolveLit` 0→3, `hidden` 0→1,
62 KB → 68 KB.

bundle (940 cols / 438 hidden / 10 folders, matching source). That check matters: see the note
in the powerbi PR — `vendor-converters.sh` bundles from `build/`, not `src/`, and will vendor a
stale `build/` silently while still stamping `PROVENANCE` with the current commit.

### DIFF
--- a/.github/workflows/hygiene.yml
+++ b/.github/workflows/hygiene.yml
@@ -165,6 +165,23 @@ jobs:
         run: |
           if [ -z "${RANGE_BASE:-}" ]; then echo "no diff range to check — skipping"; exit 0; fi
           bash tools/check-converter-provenance.sh "$RANGE_BASE" "$RANGE_HEAD"
+      - name: Converter-freshness guard (vendored-bundle staleness by age)
+        # Standing gate, NOT diff-scoped (unlike the pairing guard above, this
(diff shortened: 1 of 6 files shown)


twells89:

PR #584, opened 2026-07-31
chore(thoughtspot): re-vendor converter — literal masking + lookup declutter (mcp#120, #121)

twells89:
Bundle was pinned at `867c669`, so this picks up **more than today's fixes**.

**mcp#120** stops ThoughtSpot formula scans from treating string-literal text as live syntax.
The demonstrated defect: a literal containing an aggregate name **leaked raw sentinel bytes
into the emitted formula**, because two independent nested mask cycles corrupted each other.
The RLS path (`tsRlsExprToSigma`) now goes through the same mask, so a worksheet rule over a
bare identifier no longer ships structurally invalid.

Worth noting the review history here: a first attempt at this fix double-bracketed every
bare-identifier aggregate — `sum(net_revenue)` became `Sum([[Net] [Revenue]])`, splitting one
display name into two refs — and **all 56 regression-corpus cases passed anyway**, one of them
breaking its own fixture while still reporting PASS. A `noDoubledBracketRefs` assertion was
added upstream so the corpus can catch that class.

**mcp#121** hides the derived View's related lookup columns and folds them one folder per
relationship target, dropping none.

Markers in the regenerated bundle: `tsMaskLiterals` 0→2, `hidden` 0→1.

bundle. See the powerbi PR for why that matters: `vendor-converters.sh` bundles from `build/`,
not `src/`, and vendors a stale `build/` silently while still stamping `PROVENANCE` with the
current commit.

### DIFF
--- a/.github/workflows/hygiene.yml
+++ b/.github/workflows/hygiene.yml
@@ -165,6 +165,23 @@ jobs:
         run: |
           if [ -z "${RANGE_BASE:-}" ]; then echo "no diff range to check — skipping"; exit 0; fi
           bash tools/check-converter-provenance.sh "$RANGE_BASE" "$RANGE_HEAD"
+      - name: Converter-freshness guard (vendored-bundle staleness by age)
+        # Standing gate, NOT diff-scoped (unlike the pairing guard above, this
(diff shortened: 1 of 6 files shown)


twells89:

PR #585, opened 2026-07-31
chore(looker): re-vendor converter — LookML literal masking (mcp#120)

twells89:
Bundle was pinned at `867c669`, so this picks up **more than today's fixes**.

**mcp#120** stops the LookML cross-element calc-column pass from reading string-literal text as
live syntax. `lookml.ts` had **never been audited** for this defect class — it was the one
acknowledged unknown — and turned out to be **exposed and dangerous**: a quoted label
containing a bracketed name had that name read as a live column reference, silently relocating
or, where no derived element existed, **dropping** a fully self-contained column. A second
instance corrupted literal text inside a genuinely cross-element formula. Both demonstrated
live before any code changed.

`dbt.ts` was found to contain byte-identical copy-pasted vulnerable code, which is how the
class spread; it has no plugin bundle, so it reaches users only via the hosted MCP and CLI.

**mcp#121 does not apply here.** `lookml.ts` carries its own local copy of
`buildDerivedElements` rather than the shared one, so `hidden` is correctly **absent** from
this bundle — not a failed vendoring. That local copy is recorded as an unpatched follow-up
upstream.

Marker: `maskFormulaStringLiterals` 0→4.

bundle. See the powerbi PR: `vendor-converters.sh` bundles from `build/`, not `src/`, and
vendors a stale `build/` silently while still stamping `PROVENANCE` with the current commit.

### DIFF
--- a/.github/workflows/hygiene.yml
+++ b/.github/workflows/hygiene.yml
@@ -165,6 +165,23 @@ jobs:
         run: |
           if [ -z "${RANGE_BASE:-}" ]; then echo "no diff range to check — skipping"; exit 0; fi
           bash tools/check-converter-provenance.sh "$RANGE_BASE" "$RANGE_HEAD"
+      - name: Converter-freshness guard (vendored-bundle staleness by age)
+        # Standing gate, NOT diff-scoped (unlike the pairing guard above, this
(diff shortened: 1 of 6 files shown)


twells89:

PR #509, opened 2026-07-27
Wave 1: single-invocation runs, evidence dedup substrate, context diet, intake triage

jflores1013:
Wave 1 of the ratified speed & success program (RECONCILED-PROGRAM.md): the turn-architecture and substrate wave. **Rebased onto post-#508 main** per review; review response in the latest two commits.

Six commits:
- **Single-invocation runs** — the guaranteed Phase-1d abort becomes a bounded wait-gate; finalize chains in-process when no agent-mediated actuals remain; gap-scan/decisions/consent/cost-advisory collapse into one checkpoint with a single re-entry; mission scope threads end-to-end (scoped runs filter attributable out-of-scope gaps, fail-open on unattributed); decisions.jsonl; --quiet machine stdout.
- **Evidence ledger + version-keyed dedup + kind-parity** — append-only per-gate evidence ledger (the factory punch-list substrate); one export cache keyed (workbook, documentVersion, element) with raw-payload-only reuse (verdicts always recomputed); pooled readback; kind-parity gate.
- **Intake triage + probe litter + phase metrics** — migration-plan.json consumed when present; probe registry with crash-safe cleanup + registry-first sweep; local per-phase wall-clock capture (consent still gates any send).
- **SKILL.md context diet** — tableau SKILL.md 40,960B → 19,973B (**correction from earlier body text**: the "79KB" figure was the pre-#443 baseline; the reproducible diff is 40,960→19,973, currently 20,270B with review fixes) plus the ~227KB 13-ref mandatory pre-read reduced to a 1-file, phase-scoped contract; byte budgets CI-enforced with a ratchet baseline; gate catalog relocated set-identically (25/25).
- **Dual-review fixes (A1–A11)** — from the independent Opus-5 + Fable-5 review pass.
- **Review response** — rebase onto #508 (resolves the `list_entries` crash class), CI allow-list for the non-globbed tests, consent file-gated on **every** path (a --yes run that never surfaced the question now records an honest no-response and the sender suppresses), UTF-16-safe scope banner, domo pre-read phase-scoped to 3 refs (the new byte-budget ratchet refused to baseline a new skill — working as designed).

**On single-invocation honesty:** the reviewer's live E2E measured 5 orchestrator re-entries, not one pass. The wave's items remove the *guaranteed* re-invocations (1d abort, unconditional finalize split); they do not yet collapse the repair/verification loop — that's the tiering/factory work next. A clean cold-run re-measure is queued before we lean on the ≈52–58-turn projection.

[...]

PR #587, opened 2026-07-31
Wave 2: tiered single-invocation factory runs, evidence-bound verdicts, pooled probes, converter calc-core + object-model wiring — measured, with field fixes (1.4.2 → 1.6.0)

jflores1013:
This is the integrated Wave-2 PR: 8 file-disjoint lanes implemented, cross-reviewed, fix-passed, and merged on one integration branch, plus the field-driven fixes that landed on top of the lanes during live measurement. `tableau-to-sigma` lands at **1.6.0** (main is at 1.4.2), with twin patch bumps at integration per the version-gate discipline. 85 commits, 311 files (+47.9k/−5.6k — the bulk is corpus fixtures, tests, and generated artifacts). The branch has been kept current with `main` by merge throughout — the maintainer's parallel work through #581 (fleet pagination/lint #560–#567, converter re-vendors #569/#574/#576, domo Tracks B/C, hex-to-sigma #580) is merged forward, not duplicated, including a post-#580 `sync-shared.rb` fan-out so the new plugin's twins match the wave's canonicals.

### What the wave delivers, by lane

**1. Single-invocation runs (re-entry killers).** The three named causes of the measured 5-re-entry loop are dead: `[Metrics/<name>]` refs now validate against the DM's own metrics census (miss = hard error, `[Bogus/X]` still rejected); join-probe SQL identifiers strip role-disambiguation parens at derivation and pass one legality oracle at emission (a still-illegal key becomes a clean recorded `error` routed to the resolve path — never guessed into live SQL); `--skip-anchors-gate` is declared and forwarded like its siblings, with the gate's whole waiver-flag table audited. On top: a `--wait[=SECONDS]` foreground contract (inner exit codes pass through verbatim; exit 26 = budget exhausted, run still alive), a pass-1-tail visual-verdict wait-gate so the cold chain can actually finalize in one invocation, render-once per document version, and Tier-S checkpoint auto-defaults (every auto-answer is a ledgered decision).

**2. Tiering + factory mode.** New pure-function tier detection (`lib/tier.rb`, mechanical inputs only, fail-closed to the full battery) drives `--tier {auto|S|M|full}`. Tiers shrink budgets and duplicate oracles — never the gate catalog (25-gate set identity frozen). Factory mode is the new default: one pass + measured parity + a punch list of copy-pasteable re-entry commands rendered from the shipped degradation/evidence ledgers; `--certified` opts back into loop-to-green with the mandatory fresh-context verifier. A PARTIAL verdict with an empty punch list is itself a test failure.

**3. Evidence + verdicts (gates).** The gate half of tiering (tier read from run state, Tier-S waiver-budget scaling, GT-trio skip carried by gate 18's own valued-anchors oracle — the gate still evaluates and can fail); verifier made optional-and-labeled for factory Tier-S with the distinct greppable verdict `GREEN (factory, self-attested)` plus the matching doctrine amendment (code and doctrine no longer contradict); evidence-bound attestation and per-entry gate-16 emitted binding from the review fix-pass; the wave's single vocabulary commit for new offramp/decision tokens; the domo ledger exemption ratified as a standing decision instead of an accident.

**4. Probe/export pooling.** The probe registry moved into the export pool *before* anything was wired through it (litter red line), then ground-truth runs route through one pooled probe workbook — 1 POST + T exports + 1 DELETE, serial path kept as `--no-pool` fallback — and control flip-exports pool wide instead of K×2 serial round-trips. Registry reads zero-open after sweep on every live run recorded below.

**5. Converter calc-core.** One generated translation table (`functions.json`) replaces the hand-maintained coverage manifest and pins converter ↔ table ↔ live classifier catalog with a drift test; batched validation probes (N formulas → one probe workbook, one readback); a 10-pair differential dialect corpus as the floor under any future engine change; union support with a named-refusal stopgap (a unioned datasource can no longer produce nothing *silently*); true parameter defaults read from the workbook's own current value instead of a hardcode. All converter changes follow the vendoring discipline: patch source, rebuild, `PROVENANCE.json` `local_patches` entry in the same diff range.

**6. Object-model ("noodle") hardening.** Disconnected-table models are now a first-class detected shape: gap-scan ❌ class + checkpoint stop; evidence-ranked fact election with edge orientation and an explicit `--fact-table` override; named relationship-gap recording; structural entitlement-table RLS detection routed through the existing Port/Customize/Skip flow (planned loudly, never auto-applied); slash-in-column-name handling; `check-sql-idents` promoted to a real pre-POST gate (exit 20) and `lod_audit` FROM/ownership validation — the wrong-FROM class can no longer pass late gates. Two field-failure twin fixtures promoted into the corpus with DM goldens.

**7. Discovery scoping + bounded fetches.** `--dashboard` missions fetch member-sheet CSVs only (membership resolved at t≈0, fail-open to all-CSV with a stated reason); PNG export gets a dedicated worker from t≈0; workbook-content downloads are budget-bounded with size pre-abort — the observed multi-hour trickle wedge is capped.

**8. Measurement machinery.** `phase_metrics` captures turns/invocations/run stats mechanically; `estimate-cost.rb --from-metrics` re-fits bands from measured workdirs and refuses low-n publication; `measure-cold-run.rb` is the live cold-run harness with a terminal-only exit gate; `refs/performance.md` gains the measurement section. Numbers below come from this machinery, not hand-tallies.

### Field-driven fixes landed on top (found by running the thing live)

- **Same-named metric/column collision (F4):** the DM readback drops governed metrics that share a name with a column, which deterministically failed the workbook ref-gate on every cold run of that shape. Fixed by inlining measures on collision elements; ref-gate holds 101/101.
- **Harness resume contract (F5):** the measurement harness resumed exit-12 runs with `--finalize` but not `--actuals`, so the documented two-invocation contract could never complete. Fixed; the re-entry chain now completes correctly under the harness.
- **Role-played date dimensions:** one DM element instance *per role* even when several roles map to one physical table — role-played dims no longer collapse onto a shared element.
- **Columns pagination (prospect report):** `discover-columns` follows `nextPage` — wide tables no longer truncate at 50 columns; merged forward with the maintainer's fleet-wide pagination + lint work (#560/#565/#567) rather than duplicating it.
- **Element-filter `id` (prospect report):** phase-5 element filters carry the `id` field the Sigma spec schema requires.
- **vds-to-cdw landing collisions:** distinct captions that sanitize to the same column name no longer produce duplicate-column DDL errors or silent column loss; deterministic collision suffixes plus an ordinal caption→column provenance manifest on both warehouse targets (no-collision output stays byte-identical).

### Measurement (the wave's exit gate, run live on this head)

The wave's contract is that the published number is the *measured* one. The cold-run harness executed the full protocol on the integration head: factory default, `--tier auto`, no skip flags, no waivers, no tier overrides; every stop attributed by name; litter verified zero-open after every run.

Headline run #1 (internal reference workbook, auto-tiered **M** by the detector's own predicate — reported as the M band, not laundered into S): **4.02 min wall**, 19 turns, 2 invocations (pass 1 + the documented finalize re-entry), **strict parity PASS 5/5 exportable tiles, value_parity_score 1.0**, factory verdict PARTIAL with a 3-item punch list (the ratified factory-mode headline shape: the three items are honest scope-cuts — one source-side-unexportable tile, one unsupported toggle control, one parameter needing wiring — each with its re-entry command). Earlier diagnostic runs on the same head found, falsified, and then fixed the root causes recorded above; the parity-collapse class they exposed is verified dead end-to-end.

Full campaign numbers (measured live, 2026-07-30 → 2026-07-31, on this head's integration arc):

- Median cold-run wall: **3.09 min** (n=3 qualifying headline runs; operator clock, intake→stop; metrics-wall median 3.06, cross-check delta 0.02 on every run)
- Published band (measured, honest M tier): **2.83–4.02 min wall · 19 turns · 2 invocations (pass 1 + the documented finalize re-entry) · 1 attributed stop**, parity strict **5/5 @ value_parity_score 1.0 on all three runs**, each on a bench proven same-day fresh. The harness exit-gate's terminal-only refusal (`refused-non-terminal` — a deliberate factory-PARTIAL stop is not a terminal wall under its R2-era §5 contract) is quoted verbatim in the measurement report and disclosed rather than papered over; the band publishes under the product owner's recorded factory-PARTIAL ratification (2026-07-30).
- Per-run table (wall / turns / invocations / stops attributed / parity / verdict):

  | Run | Wall (min) | Turns | Inv. | Stop (attributed) | Parity | Verdict |
  |---|---|---|---|---|---|---|
  | headline #1 | 4.02 | 19 | 2 | census gate: 1 source-side-unexportable trend zone (the ratified scope-cut) | 5/5 @ 1.0 | factory PARTIAL, punchlist 3 |
  | headline #2 | 3.09 | 19 | 2 | census gate (same zone) | 5/5 @ 1.0 | factory PARTIAL, punchlist 3 |
  | headline #3 | 2.85 | 19 | 2 | census gate (same zone) | 5/5 @ 1.0 | factory PARTIAL, punchlist 3 |
  | re-entry proof (star-schema reference) | — | — | 2 | structural scope-cut (one parameter needs-wiring) | 9/9 @ 1.0, census 9/9 | factory PARTIAL |
  | certified (`--certified`, `--allow-missing-tiles 1` disclosed + punchlisted) | 2.34 | 19 | 2 | scope-cuts + recorded visual divergence | 5/5 @ 1.0, 3 controls flip-proven live | PARTIAL+YELLOW |

  Two diagnostic runs (parity-collapse class, root-caused to the reference environment's cross-role date views and repaired at the data layer) and one drift-voided run (bench freshness; speed number voided per fidelity-outranks-speed) precede these in the ledger and are reported honestly in the measurement doc — nothing laundered.
- Re-entry proof vs. the 5-re-entry baseline: **induced exit-4-class stop → exactly 1 re-entry (39 s) to the factory finish** — the ref-gate named the corrupted ref (1 of 102), stopped pre-POST with zero litter; the fixed spec re-entered once via the documented `--reuse-dm` + `--wb-spec` fast path, re-gated 102/102, reused the same workbook, and chained finalize in-process. Harness-recorded `re_entries = 1` on every completed run of the campaign. The first blind-graded visual pass also ran on both references (fresh context-free grader, hash-bound): chart families 10/10 identical and all graded values exact on both, with the polish gaps (palette mapping, axis-label retention, number formats, tile proportions) recorded as DIVERGENT verdicts, punchlisted, and queued as wave-3's top converter items.

### Supersedes PR #509

This PR supersedes **#509**. Its evidence-ledger substrate is carried here in canonical form: the ledger content lives in the canonical shared gate (`shared/scripts/assert-phase6-ran.rb`) and shared libs, registered in `shared/manifest.json`, with the full twin fan-out synced — `check-shared.rb` green (634/634, 1 allowlisted exception) — where #509 had edited per-plugin copies that the shared-sync machinery would have silently overwritten. Nothing from #509 is dropped; it is re-homed where the governance hooks protect it. #509 can close when this merges (closure comment drafted below).

### Reviewer's guide (suggested read order)

1. **`plugins/tableau-to-sigma/skills/tableau-to-sigma/SKILL.md` diff first.** It is the operator-facing contract for everything else: the `--wait` driving pattern, tier/factory/`--certified` semantics, the punch-list shape, and the exit-code table. If the SKILL.md story reads right, the code below is implementing it.
2. **`shared/` canonical changes + `shared/manifest.json`.** The substantive gate work is all in the canonical files: `assert-phase6-ran.rb` (tier read, waiver budgets, gate-18 oracle, evidence binding), `export_pool.rb` (registry-in-pool), `offramp.rb` (the one vocabulary commit), `evidence_ledger.rb`, `phase_metrics.rb`. Review these once, canonically.
3. **Converter: `PROVENANCE.json` `local_patches` first, then the source patches.** Every bundle change is paired with a `local_patches` entry in the same diff range (`check-converter-provenance.sh` enforces it). The `.mjs` bundle diffs are build artifacts of the source patches — verify the pairing, don't line-review the bundle.
4. **The gates last, as the proof:** `check-shared.rb` 634/634 (1 allowlisted exception), version-bump gate vs `main`, hygiene sweep, corpus/lint suites, and the six-trajectory test matrices that ship with every new guard (each new assertion has been watched failing).

**Mechanical vs. substantive:** the twin fan-out across the sibling converters (sync-shared output + per-plugin patch bumps), the regenerated `functions.json` + re-baselined corpus goldens (CI-diffed against the generator), and the version bumps are mechanical — spot-check, don't line-read. The substantive surface is: canonical `shared/` files, `migrate-tableau.rb` (lane-A orchestrator: tier, factory, wait contract, chain gate), `assert-phase6-ran.rb` (lane-B gate half), the converter source patches, and the six field-driven fixes listed above.


jflores1013:
**CI fix-pass — all 10 checks green at `cd426071`.** The run at PR-open (`e9dda6a4`) had 3 red jobs hiding 5 distinct failures; each is fixed at the root rather than suppressed:

**1. Stale Sigma test stubs** (full ruby suite — 3 crashes). Main's fleet columns-pagination (#560/#565) moved `validate-sigma-formula.rb` and `probe-controls.rb` to `Sigma.list_entries`, but three hand-rolled stub modules (total replacements via `$LOADED_FEATURES` suppression) still defined only `Sigma.request` — the stubbed subprocess died with a NoMethodError and empty stdout, surfacing as bare `JSON.parse('')` crashes. Each stub now defines `list_entries` (the cleanup test's fix went into the canonical `shared/scripts` copy and fanned out to both twins). One real bug fell out: the singleton path in `validate-sigma-formula.rb` graded an *empty* columns readback `ok` once `list_entries`' defensive `[]` replaced the old accidental crash — a vacuous pass. It now refuses loudly (the just-POSTed spec always carries the probe column), preserving the pinned nonzero-exit + at_exit-cleanup contract.

**2. Converter backslash corruption** (translation-table determinism gate). The `d839036` re-vendor (#574 — upstream string-literal masking, mcp PR #118) unescapes **every** `\X` pair in a masked literal, so `'(\d+)'` emitted `"(d+)"` and `'C:\path\file'` emitted `"C:pathfile"` — silently wrong regexes/paths in every converted calc. Main has no gate that would catch this; the branch's W2.13 freshness gate tripped because the ratified `refs/functions.json` pins the correct templates. In-place bundle patch (PROVENANCE `local_patches` entry + TASK 8 upstreaming task): quote-only unescape + no backslash-doubling on emission. An 11-case probe battery across the pre-#574, unpatched, and patched bundles confirms every PR #118 improvement is retained (THEN/ELSE/IN-in-literal inertness; proper `\"` escaping where the pre-#574 bundle emitted malformed output) while backslash content round-trips verbatim. The regenerated table differs from the ratified one **only** in `bundle_sha256`. ⚠️ *Upstream heads-up: the defect ships on the converter repo's main today — every un-patched consumer emits corrupted literals. TASK 8 in PROVENANCE.json has the port-ready details.*

**3. Windows corpus "drift"** (windows job). `corpus_check.py normalize` wrote its output file in Python text mode → CRLF on Windows vs LF-pinned goldens → byte-exact `cmp` failed with whole-file diffs. One-liner: `newline="\n"`. Also de-mines the documented golden-regeneration recipe, which on Windows would have committed CRLF goldens into an LF-mandated tree.

**4. Closure ratchet** (offline job). `probe_registry.rb` ships as two deliberately-forked per-tool copies (each binds its own `<TOOL>_TO_SIGMA_HOME` dotdir — the same convention as `learned-rules.rb`, the existing precedent). Ratified in `PER_TOOL_BY_DESIGN`; the backlog list is untouched.

**5. Sisense metric/column collision** (offline job, second push). A new *exposure*, not a regression: the workflow broadening that arrived with the merge selected `test_metric_reference.py` variants that had never run on this branch. Sisense's JAQL harvest emitted metrics named identically to fact columns; wave-2's field-verified collision rule (a collision-shaped element's metrics are dropped wholesale by the live readback) then rightly refused every governed binding. Sisense's attach loop now never creates the shape — duplicate aliases skipped, colliding metrics withheld with a flag (safe inline fallback). All six sisense tests plus the gooddata/microstrategy/thoughtspot metric-reference tests pass.

Also merged current main (#586/#588/#591) — textually clean, shared-lib drift gate unchanged at 634/634, and #586's re-vendor clears the converter-freshness staleness. Version bumps: tableau-to-sigma 1.6.1, sisense-to-sigma 1.1.2.

PR #589, opened 2026-07-31
tools: a bundle pinned by an OPEN local_patches entry warns instead of failing

twells89:
## The incoherence

The converter-freshness gate hard-failed a stale bundle while its own message said
*"do not just re-vendor blind"* — because that bundle carries patches not yet landed
upstream. **It failed you for not taking the action it warned against**, and offered no
path to green.

## Found the hard way, not reasoned

Re-vendoring `quicksight-to-sigma` (stale at `867c669`, 39 days) **destroyed a feature**:
unpartitioned `rank` / `denseRank` / `percentileRank` / `percentOfTotal` all fell to `Null`,
breaking `scripts/test-window-native.rb`.

Investigated instead of adjusting the test — and upstream `src/quicksight.ts` lists those
functions in its *cannot-be-lowered* set, with `git log -S denseRank` showing a single
commit, the original converter. **The native window lowering exists only in the vendored
bundle**, as an undocumented hand-patch with no `local_patches` ledger to warn anyone.

So that bundle's staleness was *protecting* a real feature, and the gate's advice would
have deleted it.

## The change

A stale bundle with an **OPEN** `local_patches` entry now emits `::warning` and names the
real remedy — port the entry upstream, retire it, then re-vendor.

**Teeth preserved, proven both ways:**
- stale with **no** `local_patches` → still hard-fails (the case the gate exists for)
- stale whose entries are **all SUPERSEDED** → still hard-fails, because a superseded patch
  has landed and no longer pins anything. Without this, marking every entry superseded
  would silently disarm the gate.

Three existing assertions encoded the old contract and are replaced; six new ones were
negative-proved against the old guard (all six fail), and the old assertions fail against
the new guard. `tools/test-converter-provenance.sh` is ALL PASS.

## Disclosure

This unblocks re-vendor work of mine that the gate was holding. The merits stand
independently — a gate demanding an action it simultaneously warns against is broken
regardless of who it inconveniences — but it is a gate change made by the same author whose
PR it unblocks, so it deserves a reviewer's eye rather than a rubber stamp.

## Follow-up

`quicksight`'s undocumented patch is now recorded in a `local_patches` ledger (separate PR).
The mapping should be ported upstream to `src/quicksight.ts`, then the entry retired and the
bundle re-vendored cleanly.

### DIFF
--- a/.github/workflows/hygiene.yml
+++ b/.github/workflows/hygiene.yml
@@ -165,6 +165,23 @@ jobs:
         run: |
           if [ -z "${RANGE_BASE:-}" ]; then echo "no diff range to check — skipping"; exit 0; fi
           bash tools/check-converter-provenance.sh "$RANGE_BASE" "$RANGE_HEAD"
+      - name: Converter-freshness guard (vendored-bundle staleness by age)
+        # Standing gate, NOT diff-scoped (unlike the pairing guard above, this
+        # runs every time regardless of RANGE_BASE): a merge to
+        # sigma-data-model-mcp changes nothing for any operator until someone
+        # runs tools/vendor-converters.sh and commits the regenerated bundle,
+        # and pre-existing drift nobody re-vendored in a long time has no diff
+        # to catch it on. CI cannot assume network access to that private repo,
+        # so this grades staleness by the AGE of each PROVENANCE.json's own
+        # source_commit_date (default threshold 14d, CONVERTER_STALENESS_DAYS)
+        # rather than asking upstream what's new — see the header comment in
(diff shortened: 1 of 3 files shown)


twells89:
Folded into #586. The two PRs were circularly blocked: the gate fix alone leaves main's four stale bundles hard-failing (they carry no local_patches), and the re-vendors alone leave quicksight documented-but-stale, which the old gate still failed. Neither could go green independently, so they land atomically. Governance normally separates tooling from plugins; a circular CI dependency is the case where that has to bend.

PR #596, opened 2026-08-03
fix(tableau): deny unique-on-right non-keys from join-key inference (1.6.3)

jflores1013:
Closes the residual hole you disclosed in handoff §3 and scoped in §6d — filing note: you asked for a bead with this one; there's no `bd` CLI on this machine, so please attach/file at review and I'll reference it in a follow-up commit message if needed.

**The hole:** the PR2a inference ladder's `_ID/_KEY` suffix leg is entity-agnostic by design, so a column shared on both sides that is **unique on the right but isn't the key** — `EXTERNAL_ID`, `ROW_ID`, `GUID`, `HASH_KEY` — passes both matcher legs *and* gate 16 (the probe proves uniqueness, not correctness, exactly as you corrected three times). Result: a wired join that silently returns wrong rows behind green.

**The change** (in-place bundle patch, `local_patches` entry paired in the same diff; no re-vendor — the open patches stand): `INFERENCE_KEY_DENYLIST_RE` = `(^|_)(EXTERNAL_ID|ROW_ID|GUID|HASH_KEY)$`, applied **only to inference candidacy** inside `inferRelationshipKeyByName` — a key Tableau explicitly serialized is honored untouched. A denied-only match records `unwired` with a reason naming the denied column(s) and the manual-authoring remedy, per the ledger discipline.

**Deliberate second effect worth your eye:** a denied name no longer manufactures false ambiguity. Previously `CUSTOMER_KEY` + `EXTERNAL_ID` shared on both sides = 2 key-shaped candidates → refuse-to-guess; now the genuine key wires alone. The deny-list both blocks a wrong wire and rescues a right one.

**Tests** (watched failing first — and the failures demonstrated both live bugs: the wrong wire happening via name-inference, and the genuine key lost to ambiguity): two scenarios on surgically-modified `logical-model-objectgraph` fixture variants in `test-relationship-derivation.rb`; all 17 existing assertions unchanged. `refs/functions.json` regenerated — `bundle_sha256` only, all 187 templates byte-identical (formula translation untouched).

Plugin-only change, version 1.6.3 (1.6.2 is staged on the open ruzs PR #595 — strictly increasing in either merge order). Local: corpus 25/25, translation-table, provenance + version gates, hygiene, full tableau glob green.

### DIFF
--- a/plugins/tableau-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/tableau-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "tableau-to-sigma",
   "displayName": "Tableau → Sigma",
-  "version": "1.6.1",
+  "version": "1.6.3",
   "description": "Migrate Tableau workbooks and datasources to Sigma — discovery, calc/LOD translation, data model + workbook build, and parity verification against the source warehouse. Includes a tenant assessment skill and a VDS→warehouse data-landing skill (Snowflake or Databricks).",
   "author": { "name": "Thomas Wells" },
   "license": "MIT",
--- a/plugins/tableau-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/tableau-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "tableau-to-sigma",
   "displayName": "Tableau → Sigma",
-  "version": "1.6.3",
+  "version": "1.6.5",
   "description": "Migrate Tableau workbooks and datasources to Sigma — discovery, calc/LOD translation, data model + workbook build, and parity verification against the source warehouse. Includes a tenant assessment skill and a VDS→warehouse data-landing skill (Snowflake or Databricks).",
   "author": { "name": "Thomas Wells" },
   "license": "MIT",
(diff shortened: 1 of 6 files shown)


jflores1013:
@twells89 heads-up before you burn another re-trigger: the sweep red here (and on main at `1f9d8b77`) is **deterministic, not stale** — retried 3× across two heads, fails every time on the runner while the identical tree passes locally.

Root cause: the range self-test's unresolvable-range fixture builds a shallow clone via `git clone --depth 1 file://…`, and the runner's git **2.54** silently refuses that combination (local 2.50 honors it). The clone never materializes, and the two downstream `cd`-into-missing-dir errors read as bogus behavioral failures ("refusal is loud" / "push-only silent skip"). Not an identifier hit — local `hygiene-sweep.sh` and both self-tests are clean.

Just pushed `b020416b` on top of your re-trigger: the fixture is now built with pure plumbing (`cat-file | hash-object` the tip commit, set ref, mark shallow — same shape, same SHA, zero transport), plus a hard sanity guard so a future fixture-setup failure reports as a loud fixture error instead of misleading assertion failures. Both self-tests ALL PASS locally; this head's CI run is the proof for the runner.

Same fix will clear main's red sweep once this merges. Merge plan (version-stamp monotonicity): this PR (tableau 1.6.5) then #598 (1.6.6); #595 already landed as 1.6.4.

jflores1013:
Superseded by #600 — this branch's scanned range permanently includes main's `b7cae31d` squash message (GitHub auto co-author trailer tripping the tester-identity hygiene patterns), so its commit-message guard can never pass. The tree re-lands byte-identical there; diagnosis details in the comment above.

PR #606, opened 2026-08-04
fix(domo-to-sigma): narrow IN(...) lint + drop wrong WEEKDAY rewrite (beads-sigma-kn8s, beads-sigma-nrml)

twells89:
## Summary

Two independent, already-diagnosed Beast Mode lint/normalize bugs in `convert-beast-modes.rb`, both from the [domo-to-gold Track D backlog](docs/handoff/2026-08-03-domo-to-gold-track-e-done.md).

- **beads-sigma-kn8s**: `lint_formula` blanket-banned any `IN(` substring (meant to catch raw SQL infix `x IN (a,b,c)`, which Sigma can't express), but that also rejected Sigma's own legitimate `In([col], "a", "b")` function call. Replaced the substring ban with `raw_sql_infix_in?`/`operand_immediately_before?`, which distinguishes by shape: a raw infix always has a left-hand operand touching the `IN (` keyword; Sigma's function-call form starts a fresh expression. `NOT IN` is handled as its own always-raw-infix case (there's no Sigma spelling of a two-word `NOT IN` operator), while still correctly leaving genuine `not In(...)` boolean-prefix-negation alone — this distinction was missed in an earlier draft and caught by review (see Review history below).
- **beads-sigma-nrml**: `normalize_bm` used to rewrite `WEEKDAY(` → `DAYOFWEEK(`, which broke conversion (`WEEKDAY` converts cleanly to Sigma's `Weekday()`; `DAYOFWEEK` doesn't map to anything). Verified against Domo Community Forum threads that Domo's own docs are wrong here — `WEEKDAY()` actually executes as `DAYOFWEEK()` at runtime (1=Sunday), matching Sigma's `Weekday()` exactly — so there's no real numbering mismatch to compensate for. Dropped the rewrite, removed `WEEKDAY` from the `UNSUPPORTED` list, added a non-blocking informational warning about one residual, unverifiable-at-conversion-time caveat (first-day-of-week/timezone config), and corrected stale claims in `refs/beast-mode-to-sigma.md` + `SKILL.md`.

## Review history

An independent review caught a **Critical regression** in the first cut of the `kn8s` fix: putting `not` unconditionally in the "non-operand keyword" list meant `[Region] NOT IN ("Test","Internal")` was no longer flagged as raw infix — worse than the bug being fixed, since `NOT IN` is arguably more common than plain `IN` in real Beast Modes. Fixed by special-casing `NOT IN` (recurse past the `not` to check whether *it* has an operand before it) and independently re-verified by hand-tracing the logic against 8 concrete cases (`NOT IN`, `not In(...)`, `and not In(...)`, bracketed identifiers like `[Case]`/`[Category]` immediately before `IN(`, etc.) before pushing.

## Test plan
- [x] TDD: failing tests written first for both bugs, confirmed failing pre-fix
- [x] Full plugin suite: 862 `ok:` assertions, all green (was 844 before this PR)
- [x] Accuracy fixtures (`test-convert-beast-modes-fixtures.rb`): 13/13, incl. new `D-11` (`WEEKDAY` → `Weekday`)
- [x] Corpus check (`run-corpus.sh --check domo`): 2/2, unaffected
- [x] Repo governance hooks (shared-file drift, hygiene-sweep, tree-litter, ESM imports, TWB encoding, agent-variant sync): clean
- [x] Plugin version bumped 0.10.6 → 0.10.7
- [ ] TJ review

Closes beads-sigma-kn8s, beads-sigma-nrml.

### DIFF
--- a/plugins/domo-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/domo-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "domo-to-sigma",
   "displayName": "Domo → Sigma",
-  "version": "0.10.6",
+  "version": "0.10.7",
   "description": "Migrate Domo dashboards to Sigma — DataSets → Sigma data model (flat materialized tables → table elements), Beast Mode calc fields → Sigma formulas (MySQL-dialect SQL via the shared SQL-formula converter), and cards → charts/KPIs/pivots (every card's Summary Number → a Sigma KPI, not a table). Ports page + card filters as controls, detects PDP policies for opt-in row-level security, pre-flights Domo dataset columns against the real warehouse schema before building the data model, and defers live parity verification. Bundles a read-only domo-assessment skill (public governance-dataset inventory → value/cost-ranked migration-readiness readout). Companion to the sigma-migration-skills converters.",
   "author": { "name": "Thomas Wells" },
   "license": "MIT",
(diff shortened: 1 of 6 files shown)


twells89:
Superseded — kn8s/nrml landed separately via #604 (merged) and #605 (merged), so this combined-fix branch is now redundant with main. Closing unmerged rather than merging duplicate content. Context: PR #607's handoff doc (updated 2026-08-04) has the corrected record.

PR #627, opened 2026-08-05
sigma-authoring: re-vendor sigma-workbooks OpenAPI-access fix + new spec surface (1.10.0)

twells89:
> **Depends on twells89/sigma-skills#30** — vendored from that branch (`4aa71e8`). Merge upstream first, then this. No hand edits to the vendored copy; shared files re-fanned and agent variants regenerated with this repo's own tools.

## Why this matters here

The vendored `sigma-workbooks` is the canonical spec reference every migration converter defers to, and one of its capabilities was **broken**: `wb-rep.rb capabilities` hardcoded a bare S3 URL for the compiled OpenAPI, which now returns `403 AccessDenied` because the asset requires **AWS presigned query params**.

The skill misdiagnosed this as a rotated content hash and told you to swap in a new one. The hash is unchanged — so that advice can't work, and the documented fallback (`help.sigmacomputing.com/openapi.json`) is now an HTML page that doesn't link the asset at all. Any converter that reached for `capabilities` to check a current shape was hitting a dead end.

## What's in it

**OpenAPI access**
- Presigned URL discovered at runtime from the docs site HTML (signatures carry `X-Amz-Expires=604800`, so the *procedure* is documented, not a literal link). `capabilities` verified working again — 82,535,560-byte fetch, `--kind` queries OK.
- The split specs don't document `/v2/workbooks/spec` **at all** (`sigma-rest-api.json` has zero paths matching `spec`), so the compiled asset is the only spec-side source for workbook authoring.

**New surface — live-verified 2026-08-05** (real create + `GET .../spec` readback, page render confirmed by PNG export; `/verify` alone deliberately not treated as sufficient)
- **9 previously undocumented action effects** — `open-url`, `navigate`, `select-tab` (0-based index + next/previous), `refresh-element`, `update-rows` (formula + single-row/primaryKeys), `delete-rows`, `open-document` incl. `targetControls` for deep-linking another workbook pre-filtered. Twelve effects exist; we documented three.
- **Page `type: drawer`** alongside `modal`, both via `open-overlay`/`overlayId`, plus overlay config (`width`, `position`, `showShadow`, `header`, `footer`).
- **`page-break`** element + its fixed one-grid-row height constraint.
- **Agent `greeting`** (object with `static`/`generated` modes — a bare string 400s; static confirmed rendering), `requiresApproval`, and all four `tools[]` kinds.
- Action sequences are **referenceable but not authorable** — no top-level `sequences` array, so inline `kind: effect` steps.

**Repeated containers — UI-only, and silently dropped on read**

Directly relevant to converters that round-trip specs: `GET /v2/workbooks/{id}/spec` on a workbook containing repeated containers **silently drops them**, leaving `{{[X repeated container/Col]}}` formulas that reference an element declared nowhere in the spec. A speculative `repeatFrom` passed `/verify` **and** the real create, then came back absent — `/verify` accepting a field is not evidence it exists.

So such a GET-back is **not a faithful copy** and can't be cloned to reproduce the repeat wiring.

**Masked errors** — four added to the catalog, two pointing at entirely the wrong object (an invalid `columns[]` entry reported as `document.pages[0]: Invalid type: "page"`; an unbounded layout-XML replace reported as `Duplicate layout element id`).

## Version

`sigma-authoring` **1.9.3 → 1.10.0** — new documented surface, no breaking changes. `SYNC.md` "Vendored at" SHA updated to `4aa71e8`.

## Gates

- `check-shared.rb` — OK, 676 copies match canonical
- `check-agent-variants.rb` — OK, 16 variants in sync
- `lint-skills.rb` — OK; `hygiene-sweep`, `lint-tree-litter`, ESM/encoding lints all clean
- `check-plugin-version-bump.sh origin/main HEAD` — OK

**One caveat worth flagging:** `sigma-workbooks/SKILL.md` grows 28024 → 29995 bytes (+1971). It's an existing *tracked* byte-budget exception (WARN, not FAIL), so this makes an already-over-budget file slightly larger rather than introducing a new violation. I trimmed the new section once to hold the growth down; the real fix is the pending E9 diet that moves the recipe catalog into `reference/`.


twells89:
## ⚠️ Correction re-vendored (sigma-skills `95e381f`)

The repeated-container finding in this PR's earlier commits was **wrong** and is now retracted. Re-vendored.

**Root cause:** the skill read the content-addressed Fern docs asset, which **lags the live API**. Canonical asset:

```
https://assets.sigmacomputing.com/openapi/public-rest-api/sigma-computing-public-rest-api.json
```

Plain GET, no auth/presigning/hash — safe to pin. The presigned-extraction code added earlier in this PR is deleted; it solved the wrong problem.

**`repeated-container` is a first-class element kind** — fully spec-authorable and round-trips. Verified by building one (create → `GET` → PNG: renders one card per source row) and by reading back a UI-built one intact. The `{{[X repeated container/Col]}}` bindings are correct: the name is *derived* from the source element's `name`, so it's supposed to be absent from `elements[]` — my "dangling reference" test could never have passed.

**Relevant to converters here:** `ubr5.5` (P0, "emit N GridContainers") is now *optional* rather than the only path — a native `repeated-container` is authorable directly. Bead annotated.

Also corrected: the `document` wrapper is now documented in the spec (closes the undocumented-divergence gap), `themeName`/`themeOverrides` moved to `document.settings.theme`, `container` has 6 props not 4, and two masked errors added.

Gates re-run: `check-shared` OK (680), `check-agent-variants` OK (16), `lint-skills` clean, version gate OK. Bead `beads-sigma-cofo`.

twells89:
**Holding this until the in-flight theme/re-vendor PRs land.** Not ready to merge as-is, for two reasons:

1. **Its pin is stale and now unreachable.** `SYNC.md` points at `9a671d0`, a pre-squash commit on the sigma-skills branch. twells89/sigma-skills#30 merged as squash `586524a`, so the vendor needs re-deriving against that.
2. **It would clobber concurrent work.** This PR does `rm -rf` + `cp -R` over `plugins/sigma-authoring/skills/sigma-workbooks/**`. #643 (`shared/lib/code_rep.*`, `styling.*`) and #647 (re-vendor @ `6c8ce18`) both touch files that land under that path via the fan. Re-vendoring now would revert them.

**Correct sequence:** let #643 / #644 / #647 merge, then do **one** clean re-vendor of sigma-skills `main` (currently `586524a`, which already contains #647's `6c8ce18` C1 doc *and* #30's workbook surface) — that single re-vendor supersedes this branch and #647's pin in one step. A re-vendor is cheap and belongs last.

What this PR carries that still needs to reach the plugin (all merged upstream in sigma-skills#30): the canonical OpenAPI asset fix (`wb-rep.rb capabilities` was broken outright against the presigned Fern URL), the `repeated-container` retraction, the 9 documented action effects, drawer pages, `page-break`, agent `greeting`/tool kinds, and the `document`-wrapper + `settings.theme` doc corrections.

Happy to rebuild it on request once the queue drains.

twells89:
Superseded. `main` now carries sigma-authoring **1.12.1** (this PR proposes 1.10.0), re-vendored from the canonical sigma-skills HEAD twice since this branch was cut (#660, then #677 today). Verified file-by-file: the substantive reference docs here are already on main, and merging would revert newer code_rep/styling work. Branch preserved.

PR #637, opened 2026-08-06
fix(tableau): K11, K3, K7/N1, K2, K12a — first executed tasks from the bug-register plan

twells89:
Executes Tasks 1–4 and 6 of the remediation plan. Five real converter bugs, each
with a deterministic offline regression test **proven RED on a planted defect
before the fix**. Bead: `beads-sigma-pu5e`.

| Bug | Site | Fix |
|---|---|---|
| **K11** | `migrate-tableau.rb:5283` | Phase-5b render threads called `strip` on unscrubbed `Open3` output → `Encoding::CompatibilityError` kills the thread on Windows |
| **K3** | `build-charts-from-signals.rb:8446` | `page_extras` bypassed the `seen_el_ids` namespacing; zone-id-keyed styled text collided across pages → POST fails on Duplicate id |
| **K7/N1** | `build-charts-from-signals.rb:6684` | Emitted the flat `{kind:'image', url:}` shape, rejected by the live API for **every** image; nested `source:{kind:'url'}` is required |
| **K2** | `build-workbook-spec.rb:172` | Page-id slug used a deny-list and leaked `?` `!` `#` `&` |
| **K12a** | `probe-join-keys.rb:214` | Export poll was a ~30s iteration ceiling; now wall-clock bounded via `PROBE_EXPORT_TIMEOUT_S` (default 180, env-overridable) |

Plus: plan Tasks 1–4 checkboxes ticked with an execution-status banner.

## Three things worth a reviewer's attention

**1. Two gates false-passed before they worked.** Both are documented in-code:

- The K11 test first anchored on `sigma-export-png.py` — which also appears at
  `:1412` in a help string, so the match spanned ~3900 lines and swallowed an
  unrelated scrub at `:2105`. Now anchored on the 5b thread's own `Open3` capture.
- The K7 detector took three passes: brace-balanced missed the real literal
  because of `#{...}` interpolation; line-only then false-*positived* once the fix
  wrapped the literal. Now statement-aware, with guards for flat, nested and
  wrapped-nested forms, and confirmed RED by temporarily reverting the fix.

**2. K7 inverts the original diagnosis.** The register blamed `data:` URIs. Live
probing shows `data:` URIs are fine — the *flat shape* is what fails, for hosted
URLs too. This was breaking every image, not just logos.

**3. Deliberately not touched.** The page-per-worksheet emitter also concatenates
`page_extras` but is already unique by construction (title text keyed on the
worksheet name; duplicated controls suffix both `id` and `controlId` with the
`ws_slug` for exactly this reason). The test asserts it stays untouched, so a
future "consistency" edit fails the gate.

Also: the register filed **K2** against `controlId` sanitization. Every
`controlId` path already strips punctuation; the surviving instance of that bug
class is the page-id slug.

## Verification

- 5 new tests, each with planted-defect guards.
- Full tableau suite: **233/234 pass**. The one failure, `test-calc-discovery.rb`,
  needs `TABLEAU_AUTH_TOKEN` plus a `.twb` fixture and **fails on `main` too** —
  environmental, not a regression from this branch.
- `tableau-to-sigma` bumped 1.6.10 → 1.6.11.

> ⚠️ **CI has not run on this branch.** GitHub Actions stopped firing for this repo
> around 20:38 UTC — no runs repo-wide since, across any branch, despite three
> pushes and a close/reopen here. The 7 required checks are therefore absent and
> the PR shows BLOCKED. Verification above is local. **Do not merge until Actions
> recovers and the checks go green.**

## Scope notes

- **K12** is (a) only. Sub-claims (b) `sqlproxy` right_table, (c) re-entry wiping
  probe evidence, (d) the quoted-identifier oracle each need a live run.
- Tasks 5, 8, 8.5, 8.6, 8.7, 9, 10 remain. **8.7 (K22) is a shared-lib change
  fanning out to 12 plugins** and must be its own PR with a cross-converter audit.

### DIFF
--- a/docs/superpowers/plans/2026-08-05-tableau-bug-register-remediation.md
+++ b/docs/superpowers/plans/2026-08-05-tableau-bug-register-remediation.md
@@ -2,6 +2,14 @@
 
 > **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
 
+> **Execution status (2026-08-06):** Tasks 1-4 are DONE and shipped in
+> `sigma-migration-skills` PR #637 (bead `beads-sigma-pu5e`, plugin 1.6.11). Each
+> shipped with a deterministic offline regression test proven RED on a planted
+> defect first. Full tableau suite 232/233 — the one failure,
+> `test-calc-discovery.rb`, needs `TABLEAU_AUTH_TOKEN` plus a `.twb` fixture and
+> fails on `main` too. Task 10 Step 1 (C1) shipped separately as `sigma-skills`
+> PR #33. Tasks 5-10 remain.
+
 **Goal:** Fix the `tableau-to-sigma` skill defects that survived validation against v1.6.10, and correct the authoring docs that describe API behavior which no longer exists.
 
 **Architecture:** Each task is an independent fix to one script, guarded by a deterministic offline `test-*.rb` regression test following the repo's existing convention (`ruby scripts/test-<name>.rb`, PASS/FAIL lines, non-zero exit on failure). Tasks are ordered cheapest-and-most-certain first. Two tasks (9, 10) are scoped as spec-first because they are behavioral and cannot be closed by static reasoning.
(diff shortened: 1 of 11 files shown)


twells89:
Superseded by #638, which is merged. Same commits; this branch's `pull_request` events were never delivered by GitHub (four pushes plus a close/reopen produced zero runs while other branches ran normally). Root cause of the missing checks: `corpus-check.yml` is `on: pull_request` + `push: branches:[main]`, so on a feature branch it only runs via a PR event — and those events were being dropped for this branch.

PR #641, opened 2026-08-07
fix(shared): code_rep DOC_KEYS omitted settings/agents — theme silently lost on every migration

twells89:
## The bug

`DOC_KEYS = [schemaVersion, pages, kind, layout]` drives **both** halves of the adapter — `document()` selects those keys, `metadata()` rejects them. Anything else in a flat on-disk spec is therefore dropped from `document` and pushed to the **top level** of the request body. That includes `settings` (theme + navigation) and `agents`.

## Live-probed, not inferred

Against the workbook code-rep API, 2026-08-06:

| sent as | result |
|---|---|
| `settings` at top level | `/verify` **clean**, create **200**, **theme silently gone** |
| `agents` at top level | **400** `references unknown agent` — loud, self-catching via the chat element's `agentId` |
| `document.themeName` (old in-document key) | **400** `no longer supported. Use document.settings.theme.name instead` — loud, with a perfect message |

Only the `settings` path is silent — and it's the one that matters. Every migrated dashboard comes out unthemed with nothing in any log.

This **narrows** the claim in the wrapper-migration notes that "theme and agents [are] silently dropped": agents and `document.themeName` both fail loudly. Theme-via-`settings` is the sole silent case.

## Fix

Add `settings` and `agents` to `DOC_KEYS` in all three canonical adapters (`.rb` / `.py` / `.mjs`), re-fanned to the 13 plugins that vendor them.

## End-to-end proof

Same script, both ways, against a live org — a flat spec carrying `settings.theme.name = "Dark"` pushed through the **real** adapter:

```
with fix   -> settings inside document, create 200, theme back as {"name"=>"Dark"}   ✅
reverted   -> settings at top level,    create 200, theme back as nil                ❌
```

## Tests — written first, proven to fail first

Per the repo's "prove the gate fails on a planted defect" rule: **3 failures in Ruby, 3 in Python** against the unpatched adapter; 0 after. Guards cover document placement, metadata exclusion, and rich round-trip losslessness. Re-verified by reverting the adapter under the committed tests.

## Scope / what this does NOT fix

**10+ converter scripts still emit the OLD top-level `themeName` key** (`harvest-theme-registry.py`, `verify-styling-surfaces-e2e.rb`, tableau/domo/powerbi/quicksight `build-workbook-*`, …). This adapter change does **not** rescue them — `themeName` isn't a DOC_KEY either, so it still sweeps out and is still silently ignored. Renaming those to `settings.theme.name` is cross-plugin work that belongs with the document-wrapper programme.

## Coordination

- **No open PR touches `code_rep.*`** — checked all of #610/#611/#612/#614/#618/#619/#620/#626. No conflict on the changed files.
- **`sigma-authoring` version does collide**: this bumps 1.9.3 → **1.9.4**, same target as #610/#611/#612. Per the wrapper-migration merge notes that's expected — first one merges, the rest rebase. My other PR #627 targets 1.10.0 and should merge **after** all of these.
- 13 plugins get a patch bump (user-facing bugfix, so a real bump rather than `Skip-Version-Bump`).

Gates: `check-shared` OK (680 copies), `check-agent-variants` OK (16), `lint-skills` clean, `plugin-version-bump` OK.

Bead: `beads-sigma-cofo`

### DIFF
--- a/plugins/cognos-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/cognos-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "cognos-to-sigma",
   "displayName": "Cognos → Sigma",
-  "version": "1.2.8",
+  "version": "1.2.9",
   "description": "Migrate IBM Cognos Analytics to Sigma — Data Module JSON → Sigma data model, and report-spec XML → Sigma workbook. Translates the Cognos expression DSL (total..for → window funcs, if/then/else, date/string fns) and flags what it can't (macros, running-totals, localization). Companion to the other sigma-migration-skills converters.",
   "author": { "name": "Thomas Wells" },
   "license": "MIT",
--- a/plugins/cognos-to-sigma/skills/cognos-to-sigma/scripts/lib/code_rep.mjs
+++ b/plugins/cognos-to-sigma/skills/cognos-to-sigma/scripts/lib/code_rep.mjs
@@ -3,7 +3,7 @@
 // /v2/workbooks/spec/verify); flat bodies 400. The DATA-MODEL code-rep surface is
 // NOT changing — do not use this on /v2/dataModels/.../spec payloads.
 
-export const DOC_KEYS = ['schemaVersion', 'pages', 'kind', 'layout'];
+export const DOC_KEYS = ['schemaVersion', 'pages', 'kind', 'layout', 'settings', 'agents'];
 
 const isObj = (v) => v !== null && typeof v === 'object' && !Array.isArray(v);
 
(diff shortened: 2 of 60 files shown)


twells89:
Superseded by #640, which landed the identical `DOC_KEYS` fix (adding `settings`/`agents`) while this was open — same three adapters, same root cause, and #640's test coverage (`test_settings_and_agents_stay_inside_document`, `test_settings_and_agents_round_trip_through_wrap`) is equivalent to the guards here. Nothing is lost by closing this.

Two findings from this branch worth keeping, now carried in #648:

1. **Only the `settings` path was silent.** Top-level `agents` 400s loudly (`references unknown agent`, caught by the chat element's `agentId`), and `document.themeName` 400s with an explicit "use document.settings.theme.name instead". So the silent-data-loss surface was narrower than "theme and agents silently dropped" — theme-via-`settings` was the sole silent case.
2. **`DOC_KEYS` alone doesn't fix the bug.** Every emitter was still writing the removed `themeName`/`themeOverrides` keys, so the theme was lost regardless of where the adapter put `settings`. That's #648.

PR #643, opened 2026-08-07
fix(theme): relocate themeName/themeOverrides to settings.theme.{name,overrides}

twells89:
## Why

`themeName` and `themeOverrides` were **removed from the API** — zero occurrences in the [published OpenAPI](https://assets.sigmacomputing.com/openapi/public-rest-api/sigma-computing-public-rest-api.json). The workbook theme is now `document.settings.theme.name` / `document.settings.theme.overrides`.

#640 fixed the adapter's key allowlist so `settings` round-trips. But every builder still **wrote** the removed flat pair, so the theme was dropped on every migrated workbook — silently, no error, just an unthemed dashboard.

The individual override keys (`categoricalScheme`, `colorOverrides`, `hasCards`, `borderRadius`, `elementBorder`, `titleFont`, `tableStyles`, …) are unchanged. Only the container path moved.

## Verified against the live API

Confirmed on a real end-to-end migration (org `tj-wells-1989`): POST create → GET readback returns
`document.settings.theme = {"name":"Light","overrides":{"categoricalScheme":[…]}}`, with no `themeName`/`themeOverrides` anywhere in the response.

## What changed

**Adapter** — `shared/lib/code_rep.{rb,py,mjs}`, fanned out to all 14 targets:
- `document()` folds a legacy `themeName`/`themeOverrides` artifact forward into `settings.theme`, so specs and fixtures written before the move still produce a valid body. An already-correct document is returned untouched.
- `metadata()` no longer leaks the removed keys to the top level.
- New `set_theme()` emitter helper and `theme()` reader, so shape knowledge lives in one place rather than being hand-rolled per call site.

**Emitters** (were writing the removed pair):
`tableau lib/theme_derive.rb` · `domo build-workbook-spec.rb` · `quicksight build-workbook-from-quicksight.rb` · `powerbi build-workbook-from-pbir.rb` · `shared/lib/styling.{rb,py}` · `shared/scripts/verify-styling-surfaces-e2e.rb`

**Readers** (a reader still looking for `themeOverrides` sees nothing and reports "no theme"):
`tableau lib/style_normalize.rb` · `migrate-tableau.rb` · `post-and-readback.rb` · `shared/scripts/harvest-theme-registry.py`

**Docs** — the canonical `styling.md` spec block showed the removed pair as a top-level YAML key; corrected to nested, with a note that the old pair loses the theme silently. Six other skill docs retargeted.

## Tests

7 new adapter cases per language: the fold, non-clobber of an existing nested theme, sibling-`settings` preservation, `set_theme` merge semantics, and the both-shapes reader.

**Proven RED against the pre-fix adapter** (rb: 2 failures + 3 errors; py: 1 failure + 4 errors), GREEN after. Styling goldens retargeted to the nested patch shape.

Gates: `check-shared` OK (680 copies) · `check-plugin-version-bump` OK (13 plugins bumped) · skill-lint, hygiene-sweep, tree-litter, ESM, agent-variants all clean.

## Note for a follow-up (not in this PR)

`styling.rb` is duplicated into `sigma-authoring` but is **not** in `shared/manifest.json`, so `sync-shared` doesn't manage it and the copies drift silently. I patched both by hand here; adding it to the manifest would prevent the next divergence.


twells89:
Cross-ref: closed my duplicate #648 in favour of this and #644. Measured findings from that branch (probe matrix, harvester 0→2 themes on 60 live workbooks, Sigma lowercasing hex on readback, and two testing traps) are in [my comment on #644](https://github.com/twells89/sigma-migration-skills/pull/644#issuecomment-5216358825) — the hex-lowercasing one in particular will bite any theme parity assertion.

twells89:
Superseded by #653.

Branched off `fc005b32` before #644 merged. Since this repo squash-merges, GitHub computes the diff against the *current* main tip — merging this would have reverted #644's converter-emitter work across 14 overlapping files while still showing mergeable/CLEAN.

#653 rebuilds the non-overlapping delta (adapter theme-fold + `set_theme`/`theme` helpers, `styling.{rb,py}` + the unmanaged sigma-authoring duplicate, and the canonical `styling.md` spec doc) on top of the current tip, and re-runs #644's tests green.

PR #644, opened 2026-08-07
fix(theme): move themeName/themeOverrides emission to document.settings.theme

twells89:
## Summary
- Follow-on to #640: that PR fixed the `code_rep` adapter's key allowlist, but the 4 converters that stamp a source-derived theme (domo, powerbi, quicksight, tableau) still wrote flat top-level `themeName`/`themeOverrides` — invalid on write now that themes live at `document.settings.theme.{name,overrides}`. Every migrated workbook's theme was still being silently dropped regardless of #640.
- **tableau-to-sigma**: `theme_derive.rb`'s `apply!()` (the single shared function used by both the orchestrated and standalone build paths) now emits `settings.theme`. Updated the 3 downstream in-pipeline readers of the old flat shape: `post-and-readback.rb`'s `ensure_theme!` choke point, `migrate-tableau.rb`'s summary line, `style_normalize.rb`'s SN-5 royal-scheme re-derivation.
- **domo / powerbi / quicksight**: single emission site each in their respective `build-workbook-*.rb`, same fix (quicksight's Light/Dark conditional theme-name logic preserved exactly).
- **shared/scripts/harvest-theme-registry.py**: read the live per-workbook theme name from `document.settings.theme.name` instead of the now-nonexistent top-level `themeName` — was silently reporting zero themes org-wide.
- **shared/scripts/verify-styling-surfaces-e2e.rb**: this creds-gated live probe never adopted the `document` wrapper at all (#608-#640 covered the standard per-plugin `post-and-readback.rb`/`put-layout.rb`, not this one-off E2E script) — every POST/PUT/GET in it was flat and would 400 or silently misread today's live API, independent of the theme issue. Added the `code_rep` wrap/unwrap at its POST/PUT and both GET call sites, plus the theme-shape fix, so the script is actually runnable again. (Not independently re-verified live — it's creds-gated and outside offline CI; verified by inspection + mirroring the exact pattern already proven in `post-and-readback.rb`.)
- **refs/fidelity-recipes.md** (+ `composition-recipe.md`, `fidelity-rubric.md`, `test-fidelity-recipes.rb`): tableau's hand-authored-patch recipe catalog documented the same stale flat shape — an agent following it to hand-fix a theme post-build would have hand-authored a patch that gets silently dropped too.

## Test plan
- [x] `test-theme-ensure.rb` and `test-style-normalize.rb` updated to assert the new shape; confirmed RED against the pre-fix code (2 + 6 failures respectively — reverted just the code files, kept the updated tests), GREEN after restoring the fix
- [x] `test-fidelity-recipes.rb` — ALL PASS (recipe catalog + calibration ledgers still converge with the new shape)
- [x] `test-build-workbook.rb` (domo, pre-existing suite, unrelated to theme) — ALL PASS, confirms no regression
- [x] `shared/lib/tests/test_code_rep.{rb,py}` — unaffected, still pass
- [x] `ruby -c` / `python3 -m py_compile` clean on all 4 build scripts + harvest-theme-registry.py + verify-styling-surfaces-e2e.rb
- [x] No existing test harness for the domo/powerbi/quicksight single-site emitters — verified each by isolated snippet execution reproducing the exact conditional logic (shown in review)
- [x] Pre-commit governance hooks (hygiene-sweep, byte budgets, plugin-variant sync, converter freshness) all passed at commit time


twells89:
Independently did this same sweep before spotting #643/#644 (my #648, now closed as duplicate — yours covers everything mine did, plus `refs/*.md` docs I didn't touch). Standing down rather than adding a third conflicting PR. A few measured findings from my run that may be useful here:

**The probe matrix — only one path is silent.** Live against the workbook code-rep API 2026-08-06:

| sent as | result |
|---|---|
| `settings` at top level (swept out of `document`) | `/verify` clean, create **200**, **theme silently gone** |
| `agents` at top level | **400** `references unknown agent` — loud, caught by the chat element's `agentId` |
| `document.themeName` | **400** `no longer supported. Use document.settings.theme.name instead` — loud, explicit |

So the silent-data-loss surface is narrower than "theme and agents silently dropped": theme-via-`settings` is the **sole** silent case. The other two fail loudly and are self-catching.

**The harvester bug is measurable.** `harvest-theme-registry.py` reading only the old top-level key returned `None` for every workbook and wrote an **empty registry** — indistinguishable from "this org has no themes", which silently killed the authoring skill's "use the registry instead of asking blind" guidance. On 60 live workbooks: **before 0 themes, after 2** (9× `Light`, 1 org UUID). Worth confirming your version reproduces that.

**Sigma lowercases hex on readback** (`#0E8DA0` → `#0e8da0`). Any theme parity/round-trip comparison must be case-insensitive — my first end-to-end assertion reported a false "theme LOST" on exactly this.

**Two testing traps I hit:**
- Several theme fixtures kept passing after the change because they exercised the **legacy fallback**, not the new path — the real path was uncovered until I moved the fixtures onto the new shape.
- `test-style-normalize.rb` was **exiting 1 on a stderr crash while its tail still printed `PASS`**. Worth verifying these suites by exit code rather than output.

**One thing possibly not in either PR:** a canonical `CodeRep.theme_settings(name:, overrides:)` + `merge_settings(doc, fragment)` pair (`.rb`/`.py`/`.mjs`), so the path lives in one place and a theme merge can't clobber `settings.navigation`. Happy to send that as a small standalone follow-up if you want it — say the word, otherwise I'll leave it.

PR #645, opened 2026-08-07
fix: theme moved to document.settings.theme — every converter still wrote the removed keys

twells89:
> **Stacked on [#641](https://github.com/twells89/sigma-migration-skills/pull/641)** (base = `fix/coderep-doc-keys-settings-agents`). #641 stops the adapter sweeping `settings` out of `document`; this fixes the emitters and readers that were still writing/reading the removed keys. Neither is sufficient alone.

## The bug

`themeName` / `themeOverrides` were **removed** from the workbook spec — theming is `document.settings.theme.{name,overrides}`. Live-probed 2026-08-06:

| sent as | result |
|---|---|
| `document.themeName` | **400** — `no longer supported. Use document.settings.theme.name instead` |
| top-level `themeName` | accepted, create **200**, **silently ignored** |

Every emitter in the repo still wrote the removed keys, so migrated dashboards came out **unthemed with no error anywhere**.

## Write side — was producing unthemed dashboards

- `tableau` `lib/theme_derive.rb` — the shared derive path
- `domo` `build-workbook-spec.rb`
- `powerbi` `build-workbook-from-pbir.rb`
- `quicksight` `build-workbook-from-quicksight.rb`
- `shared/lib/styling.{rb,py}` — `chart_color(categorical:)`
- `shared/scripts/verify-styling-surfaces-e2e.rb` — this one **probed** the old shape, so styling surfaces 5 and 6 would have reported a **false NO-GO**

## Read side — was finding nothing

`harvest-theme-registry.py` read only the old top-level key, returning `None` for every workbook and writing an **empty registry** — indistinguishable from "this org has no themes". The authoring skill's "use the per-org theme registry instead of asking blind" guidance was silently dead.

**Proven on 60 live workbooks:** before **0 themes**, after **2** (9× `Light`, 1 org UUID).

Tableau's `style_normalize` / `migrate-tableau` / `post-and-readback` now read through a new `ThemeDerive.theme_overrides` accessor (current shape, legacy fallback for flat artifacts still on disk) rather than hand-rolling the path.

## The path now lives in one place

`Sigma::CodeRep.theme_settings` + `merge_settings` (`.rb`/`.py`/`.mjs`) — the latter deep-merges so a theme never clobbers `settings.navigation` — plus `ThemeDerive.theme_overrides` on the read side.

## End-to-end proof

`ThemeDerive.apply!` → `code_rep` adapter → live create → `GET`:

```
READBACK settings.theme -> {"name"=>"Light",
  "overrides"=>{"colorOverrides"=>{"backgroundCanvas"=>"#ffffff"},
                "categoricalScheme"=>["#0e8da0", "#123456"]}}
✅ theme survived the full converter -> adapter -> API round trip
```

**Gotcha found while proving it (documented in `theme_derive.rb`):** Sigma **lowercases hex colors** on readback (`#0E8DA0` → `#0e8da0`). Any theme parity comparison must be case-insensitive — my first assertion reported a false "theme LOST" on exactly this.

## Tests

New guards in `test_code_rep.{rb,py}` (shape, never emits the removed keys, deep-merge preserves `navigation`) and `test-theme-derivation.rb` (`apply!` writes the new path **and never the old**; accessor reads current / `document`-wrapped / legacy / nil / garbage).

All fixtures moved off the removed keys so the suites exercise the **current** path — several were passing only via the legacy fallback, leaving the real path uncovered.

Verified **by exit code**, not by eyeballing output: `test-style-normalize` was exiting 1 on a stderr crash while its tail still printed `PASS`. All 10 suites (4 shared + 6 tableau) exit 0.

## Scope

13 plugins get a patch bump (user-facing bugfix). Gates: `check-shared` OK (680), `check-agent-variants` OK (16), `lint-skills` clean, version gate OK.

**Not covered:** `sigma-authoring`'s vendored `scripts/lib/styling.rb` still emits the old shape — it vendors from the **sigma-skills** repo, not `shared/`, so it needs the same fix upstream there and a re-vendor. Flagged on the bead.

Bead: `beads-sigma-cofo`

### DIFF
--- a/plugins/cognos-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/cognos-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "cognos-to-sigma",
   "displayName": "Cognos → Sigma",
-  "version": "1.2.8",
+  "version": "1.2.9",
   "description": "Migrate IBM Cognos Analytics to Sigma — Data Module JSON → Sigma data model, and report-spec XML → Sigma workbook. Translates the Cognos expression DSL (total..for → window funcs, if/then/else, date/string fns) and flags what it can't (macros, running-totals, localization). Companion to the other sigma-migration-skills converters.",
   "author": { "name": "Thomas Wells" },
   "license": "MIT",
--- a/plugins/cognos-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/cognos-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "cognos-to-sigma",
   "displayName": "Cognos → Sigma",
-  "version": "1.2.9",
+  "version": "1.2.10",
   "description": "Migrate IBM Cognos Analytics to Sigma — Data Module JSON → Sigma data model, and report-spec XML → Sigma workbook. Translates the Cognos expression DSL (total..for → window funcs, if/then/else, date/string fns) and flags what it can't (macros, running-totals, localization). Companion to the other sigma-migration-skills converters.",
   "author": { "name": "Thomas Wells" },
   "license": "MIT",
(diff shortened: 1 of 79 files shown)


twells89:
Superseded by #648 — same change rebuilt on current `main`.

This branch was stacked on #641, which is now closed as a duplicate of #640. Merging `main` back in produced 58 conflicted files (every fanned `code_rep.*` copy), so rebuilding was cleaner than untangling it. #648 is a clean cherry-pick with 0 deletions vs `main`, versions re-derived against the moved baseline, and all 10 test suites green by exit code.

PR #648, opened 2026-08-07
fix: theme moved to document.settings.theme — every converter still wrote the removed keys

twells89:
Rebuilt on current `main` (supersedes #645, which was stacked on #641 — see below). Cherry-picked clean; **0 deletions vs main**.

## Why this is still needed after #640

#640 landed the `DOC_KEYS` fix (adapter stops sweeping `settings`/`agents` out of `document`) — that was the same fix my #641 carried, so **#641 is closed as duplicate**. But `DOC_KEYS` alone doesn't help: **every emitter was still writing the removed `themeName`/`themeOverrides` keys**, so the theme was lost anyway. This fixes the emitters and readers.

## The bug

`themeName`/`themeOverrides` were removed from the workbook spec. Live-probed 2026-08-06:

| sent as | result |
|---|---|
| `document.themeName` | **400** — `no longer supported. Use document.settings.theme.name instead` |
| top-level `themeName` | accepted, create **200**, **silently ignored** |

## Write side — was producing unthemed dashboards

`tableau lib/theme_derive.rb` (the shared derive path) · `domo build-workbook-spec.rb` · `powerbi build-workbook-from-pbir.rb` · `quicksight build-workbook-from-quicksight.rb` · `shared/lib/styling.{rb,py}` · `shared/scripts/verify-styling-surfaces-e2e.rb` — that last one **probed** the old shape, so styling surfaces 5 and 6 would have reported a **false NO-GO**.

## Read side — was finding nothing

`harvest-theme-registry.py` read only the old top-level key → `None` for every workbook → **empty registry**, indistinguishable from "this org has no themes". The authoring skill's "use the registry instead of asking blind" guidance was silently dead.

**Measured on 60 live workbooks: before 0 themes, after 2** (9× `Light`, 1 org UUID).

Tableau's `style_normalize` / `migrate-tableau` / `post-and-readback` now read via a new `ThemeDerive.theme_overrides` accessor (current shape + legacy fallback for flat artifacts on disk).

## One place knows the path

`Sigma::CodeRep.theme_settings` + `merge_settings` (`.rb`/`.py`/`.mjs`; the latter deep-merges so a theme never clobbers `settings.navigation`), plus `ThemeDerive.theme_overrides`.

## End-to-end proof

`ThemeDerive.apply!` → adapter → live create → `GET`:

```
READBACK settings.theme -> {"name"=>"Light",
  "overrides"=>{"colorOverrides"=>{"backgroundCanvas"=>"#ffffff"},
                "categoricalScheme"=>["#0e8da0", "#123456"]}}
✅ survived the full converter -> adapter -> API round trip
```

**Gotcha documented in `theme_derive.rb`:** Sigma **lowercases hex colors** on readback (`#0E8DA0` → `#0e8da0`). Theme parity comparisons must be case-insensitive — my first assertion reported a false "theme LOST" on exactly this.

## Tests

New guards in `test_code_rep.{rb,py}` (shape; never emits the removed keys; deep-merge preserves `navigation`) and `test-theme-derivation.rb` (`apply!` writes the new path **and never the old**; accessor reads current / `document`-wrapped / legacy / nil / garbage).

All fixtures moved off the removed keys so suites exercise the **current** path — several were passing only via the legacy fallback, leaving the real path uncovered.

Verified **by exit code**: `test-style-normalize` was exiting 1 on a stderr crash while its tail still printed `PASS`. All 10 suites (4 shared + 6 tableau) exit 0.

## Scope

13 plugins get a patch bump, re-derived against current `main`. Gates: `check-shared` OK (680), `check-agent-variants` OK (16), `lint-skills` clean, version gate OK.

**Companion:** `sigma-authoring`'s vendored `styling.rb` had the same bug but vendors from **sigma-skills** (separate lineage) — fixed there in twells89/sigma-skills#30, which is merged; it reaches this repo on the next re-vendor (#627).

Bead: `beads-sigma-cofo`

### DIFF
--- a/plugins/cognos-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/cognos-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "cognos-to-sigma",
   "displayName": "Cognos → Sigma",
-  "version": "1.2.9",
+  "version": "1.2.10",
   "description": "Migrate IBM Cognos Analytics to Sigma — Data Module JSON → Sigma data model, and report-spec XML → Sigma workbook. Translates the Cognos expression DSL (total..for → window funcs, if/then/else, date/string fns) and flags what it can't (macros, running-totals, localization). Companion to the other sigma-migration-skills converters.",
   "author": { "name": "Thomas Wells" },
   "license": "MIT",
--- a/plugins/cognos-to-sigma/skills/cognos-to-sigma/scripts/lib/code_rep.mjs
+++ b/plugins/cognos-to-sigma/skills/cognos-to-sigma/scripts/lib/code_rep.mjs
@@ -26,3 +26,33 @@ export function metadata(response) {
 export function wrap(doc, extra = {}) {
   return { ...extra, document: doc };
 }
+
+// The ONE place that knows where theming lives.
+//
+// `themeName` / `themeOverrides` were REMOVED from the workbook spec. Theming is
+// now `document.settings.theme.{name,overrides}`. Live-probed 2026-08-06:
+// `document.themeName` is a hard 400 ("no longer supported. Use
+// document.settings.theme.name instead"), and a themeName that lands at the TOP
+// level (outside `document`) is accepted and SILENTLY IGNORED — the workbook is
+// created unthemed with no error. Emitters must route through here.
+//
+// Returns {} when there is nothing to say, so it is safe to merge blindly.
+export function themeSettings({ name, overrides } = {}) {
+  const theme = {};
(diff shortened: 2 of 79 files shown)


twells89:
Closing as duplicate — #643 and #644 together cover everything here (converters, `theme_derive`, harvester, e2e verifier, `shared/lib/styling`, `code_rep`) plus `refs/*.md` docs this PR didn't touch. They were opened concurrently; I didn't spot them until pre-merge checks. Merging a third overlapping PR would have conflicted both.

Measured findings from this branch posted to #644 (probe matrix showing only the `settings` path is silent, the harvester 0→2 themes measurement, the hex-lowercase readback gotcha, and two testing traps).

Branch `fix/theme-settings-path-v2` left in place in case any of it is wanted.

PR #655, opened 2026-08-07
feat(domo-to-sigma): the gate-1 parity oracle + coverage census

twells89:
Builds gate 1's missing evidence source, and closes a separate hole that let a dropped card still be GREEN. v0.14.11 → v0.16.0. Advances beads `2tkm` and (partially, for domo) `59mk`.

## The parity oracle — and a correction to how it was sized

The 2026-08-05 gold audit budgeted **3–5 days** on the premise that expected values had to be **re-derived** from `Domo.query_dataset`: reimplementing every card's aggregation, filters, rolling date window, top-N and `COUNT DISTINCT` in our own code and hoping it matched Domo. The audit catalogued the hazards itself — 28 of 36 cards carry a relative window, `COUNT DISTINCT` is a compound flag that reads as plain `COUNT` if missed, top-N has no documented tie-break.

**`GET /api/content/v1/cards/{id}/data` removes the entire class.** Domo evaluates its own card — window, filters, aggregation, ordering, limit — and returns the rendered rows, already in the `[[dim, val], …]` shape `verify-parity.rb` wants, plus a `summaryNumber` for the KPI. We re-derive nothing.

Also worth recording: `verify-parity.rb` makes **no Sigma call at all** — it is a pure JSON differ requiring *both* sides pre-collected, and `build-parity-plan.rb` emits neither. That's why this is two collectors plus a join, not one script.

| script | role |
|---|---|
| `collect-parity-expected.rb` | Domo's own rendered card values |
| `collect-parity-actuals.rb` | live Sigma element exports |
| `build-parity-oracle.rb` | join → `verify-parity` plan + exclusion ledger |
| `lib/element_export.rb` | the export+poll fetch |

**Measured live** against page `59931332`: 36/36 cards → HTTP 200, 35 carry rows, 30 a usable summary value. Mapped onto the gate's own chartable predicate (75 spec elements → 10 masters → 65 chartable = 36 base + 29 `-summary`): **64 of 65 tiles covered**. That 65/36/29 split reproduces the audit's census exactly, arrived at independently.

**Independently validated 29/29, zero divergence.** Every KPI printed on the Domo page was read off a full-page screenshot and compared to the `summaryNumber` the collector fetched — all 29 comparable cards match exactly. The API value *is* what a human sees, established without using the API to check the API.

**The one un-scoreable tile** is `el-983053598`. Three unrelated methods agree: the audit's static analysis ("the date window is baked into the plotted VALUE"), the API returning no `rows` key, and the source card rendering **"No data in filtered range"** on screen. It is excluded *with that reason*, never dropped.

## Not a port of Tableau's collector

Tableau's `collect-parity-actuals.rb` is 441 lines with only 2 Tableau-specific, so a port looked free — but most of its bulk handles problems measured on **Tableau** workbooks (wide pivot-grid exports needing a totals-JSON fallback, million-row tables that hung the pool). None observed on Domo. Domo already owned the identical fetch, live-proven since 2026-07-30, inside `verify-warehouse.rb`.

The pivot caveat is **recorded, not solved**: Domo's spec emits 13 `table` elements and whether any export wide is unknown until a live run, so the collector records each tile's column count and header shape — the first live run shows which came back wide, and those become recorded exclusions with a measured reason.

## Coverage census — a dropped card can no longer be GREEN

GREEN requires an **empty** degradation ledger, and `scope_cuts` derives from exactly two artifacts: `coverage.json`, and `parity-final.json`'s `tile_census`. Domo can never fill the second (that key is reserved for Tableau's *zone* shape; publishing anything else turns gate 5's honest SKIP into a vacuous "0 zones, 0 unmatched" — the unearned green caught on #631). And domo emitted **no** `coverage.json`: the only mentions in the skill were the two *readers*.

So a card producing no Sigma element was invisible to the ledger. `phase6-parity-domo.rb`'s census doesn't cover it either — its denominator is elements *in* the spec.

**Proven by planting a drop:** removing card `1058425328`'s elements gives census 35/36 · 1 scope-cut · verdict **PARTIAL**, against **GREEN** for an empty ledger. Unmodified run: 36/36.

`degraded` is deliberately **not** emitted — classifying free-text warning prose into severities is guesswork, and a fabricated `degraded` would cap a run for an unverifiable reason. Warnings ride at an inert severity the ledger ignores.

## Reconciled with #649

#649 landed concurrently and writes the same exclusions file, but a **complementary** class: "cannot agree by construction" (refused date window) vs this branch's "could not be collected". Two silent bugs fell out and are fixed: last-writer-wins clobbering, and a #649-excluded tile being "verified" anyway (it's still in the plan and both sides are collectable, so it would score a guaranteed DIVERGE). The generator now runs *before* the join, which honours prior exclusions over verification.

## Verification

Fixture-driven end to end, plus **fail-first proven throughout**: planting a wrong expected value drops 2/2 → 1/2; stubbing the census call fails the emit guard; flipping the oracle to opt-in fails the default-on guard. A chain that only ever passes proves nothing.

Two of my own bugs were caught by running the emitted plan through the real verifier instead of assuming its contract: `requested_columns` set to column IDs (would have crashed every tile) and companion-KPI expectations shaped `[[title, value]]` against Sigma's single-column KPI export (would have read as 29 genuine failures).

Whole domo offline suite green; 12 new static wiring guards. The one non-passing test is the pre-existing `test-telemetry-gate.rb` `LoadError` (bead `nirn`).

**Depends on #654** (the re-vendor) for the live run to clear gate 3.

### DIFF
--- a/plugins/domo-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/domo-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "domo-to-sigma",
   "displayName": "Domo \u2192 Sigma",
-  "version": "0.14.12",
+  "version": "0.15.0",
   "description": "Migrate Domo dashboards to Sigma \u2014 DataSets \u2192 Sigma data model (flat materialized tables \u2192 table elements), Beast Mode calc fields \u2192 Sigma formulas (MySQL-dialect SQL via the shared SQL-formula converter), and cards \u2192 charts/KPIs/pivots (every card's Summary Number \u2192 a Sigma KPI, not a table). Ports page + card filters as controls, detects PDP policies for opt-in row-level security, pre-flights Domo dataset columns against the real warehouse schema before building the data model, and defers live parity verification. Bundles a read-only domo-assessment skill (public governance-dataset inventory \u2192 value/cost-ranked migration-readiness readout) and domo-import-to-snowflake for landing DataSets with no connector-backed warehouse source. Companion to the sigma-migration-skills converters.",
   "author": {
     "name": "Thomas Wells"
--- a/plugins/domo-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/domo-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "domo-to-sigma",
   "displayName": "Domo \u2192 Sigma",
-  "version": "0.15.0",
+  "version": "0.16.0",
   "description": "Migrate Domo dashboards to Sigma \u2014 DataSets \u2192 Sigma data model (flat materialized tables \u2192 table elements), Beast Mode calc fields \u2192 Sigma formulas (MySQL-dialect SQL via the shared SQL-formula converter), and cards \u2192 charts/KPIs/pivots (every card's Summary Number \u2192 a Sigma KPI, not a table). Ports page + card filters as controls, detects PDP policies for opt-in row-level security, pre-flights Domo dataset columns against the real warehouse schema before building the data model, and defers live parity verification. Bundles a read-only domo-assessment skill (public governance-dataset inventory \u2192 value/cost-ranked migration-readiness readout) and domo-import-to-snowflake for landing DataSets with no connector-backed warehouse source. Companion to the sigma-migration-skills converters.",
   "author": {
     "name": "Thomas Wells"
--- a/plugins/domo-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/domo-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "domo-to-sigma",
   "displayName": "Domo \u2192 Sigma",
-  "version": "0.16.0",
+  "version": "0.16.1",
   "description": "Migrate Domo dashboards to Sigma \u2014 DataSets \u2192 Sigma data model (flat materialized tables \u2192 table elements), Beast Mode calc fields \u2192 Sigma formulas (MySQL-dialect SQL via the shared SQL-formula converter), and cards \u2192 charts/KPIs/pivots (every card's Summary Number \u2192 a Sigma KPI, not a table). Ports page + card filters as controls, detects PDP policies for opt-in row-level security, pre-flights Domo dataset columns against the real warehouse schema before building the data model, and defers live parity verification. Bundles a read-only domo-assessment skill (public governance-dataset inventory \u2192 value/cost-ranked migration-readiness readout) and domo-import-to-snowflake for landing DataSets with no connector-backed warehouse source. Companion to the sigma-migration-skills converters.",
   "author": {
     "name": "Thomas Wells"
--- a/plugins/domo-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/domo-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "domo-to-sigma",
   "displayName": "Domo \u2192 Sigma",
-  "version": "0.16.1",
+  "version": "0.16.2",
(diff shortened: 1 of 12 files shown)


twells89:
**Self-found bug, fixed in `22173e24` — this would have produced a ~45% false-failure rate on gate 1.**

Domo's card-data summary carries **both** forms, and the collector was reading the wrong one:

```json
{"label": "Sales in Period", "value": "$9.7M", "number": 9690690.9317}
```

It returned `value` — the *display* string. `verify-parity.rb` normalises numerically, and in Ruby:

| stored | `.to_f` | actual |
|---|---|---|
| `"$9.7M"` | **0.0** | 9690690.9317 |
| `"35.61%"` | 35.61 | 0.3561 (**100×**) |
| `"950.9K"` | 950.9 | 950874 |

So every companion KPI tile would have compared a mangled number against Sigma's raw export and DIVERGED — **all 29 of the 65 tiles**. Nothing about that failure points at the collector; it reads as the migration getting every KPI wrong, and costs someone a day hunting a conversion bug that doesn't exist.

Now takes `summary.number` when Numeric, falling back to display forms only when no numeric form exists at all. `status: "not_ran"` still refused — Domo declining to compute a KPI is not a zero.

Re-collected live: **30/30 summary values now numeric** — `0.566063538817589` where it stored `"56.61%"`, `3180018` for `"3.2M"`. These agree with the display forms, so the 29/29 screenshot cross-check still stands; that validated the *display* form, which was never the problem.

New `test/test-parity-oracle.rb`, 16 assertions, using this repo's own idiom for un-require-able CLIs (extract the pure function and eval it, as `test-migrate-domo.rb` does for `render_target_page`). Every payload shape is verbatim from a live response. **Fail-first verified:** removing the two-line `number` preference fails 4 of 16.

It also pins `card_id_for`'s anchoring — a master element, a `put-layout` header element, and an over-long suffix must all trace to *no* card, so they become loud exclusions rather than silent skips.

twells89:
## Adversarial review: 11 findings raised, 7 confirmed, 4 refuted — all fixed in `b4bd00a9`

Four independent reviewers converged on **one root cause**, and it's the worst class this chain can have: it produced an **unearned PASS on a genuine migration bug**. Reproduced end to end against the real element ids, before and after.

### Domo reuses generic summary labels

On the real 65-tile page, **eleven tiles share a display name** with at least one other:

| name | count |
|---|---|
| `New Visits in Period` | 4 |
| `Change over 7 Days` | 3 |
| `Surveys in Period` | 2 |
| `US Leads in Period` | 2 |

This was knowable from adjacent code — `phase6-parity-domo.rb:190-193` says it outright ("the same KPI title repeated on two pages is routine") and compares as a **multiset** for exactly this reason. I keyed by name anyway.

### Defect 1 — actuals keyed by name (critical)

Two silent failures at once: last writer wins in the thread pool, non-deterministically, and the losing exports vanished with **no `unavailable` entry** — contradicting the script's own documented guarantee. The join then attached that one surviving export to *every* tile sharing the name.

Measured, 3 tiles named `Change over 7 Days`, real element ids, expected 999/7/7 vs actuals 999/42/7:

| | result |
|---|---|
| **before** | 1 of 3 exports survived, `unavailable: []`<br>`el-503650739-summary` (999==999, a real match) → **DIVERGE**<br>`el-1136570741-summary` (42 vs 7, a **real bug**) → **PASS 100%** |
| **after** | 3 of 3 survive, each tile joined to its own → **2 PASS / 1 DIVERGE**, the real bug caught |

### Defect 2 — prior exclusions keyed by name (critical)

With one card construction-disqualified and three tiles sharing its name, **all three** were exempted from scoring — two fully collectable and never flagged. A smaller denominator reads as a cleaner pass. Now matched on `evidence.element_id`, which #649 already records. A legacy name-only entry is honoured but **consumed once** rather than broadcast: an ambiguous exclusion should under-apply, never over-apply.

### Defect 3 — Rule-0 KPIs got no shape reconciliation (important)

The single-cell reshape keyed on `-summary`, so a card whose *own* element is `kpi-chart` took the full multi-column `rows` — never matchable against Sigma's single-cell KPI export. Same bug already fixed once for the companions. Now keys on `is_summary || sigma_kind == 'kpi-chart'`.

### Verification

`test/test-parity-oracle-dupenames.rb` — 16 assertions driving the **real scripts as subprocesses** over the three real colliding element ids, with deliberately different values so a cross-wire can't hide behind coincidence. **Fail-first verified:** reverting the one-line key change fails 12 of 16.

A duplicate *element id* is a different, louder problem (the plan listed one element twice) and is now recorded in `unavailable` rather than overwritten.

One verifier hit its structured-output retry cap, so one finding went unverified — not counted among the confirmed 7.

PR #661, opened 2026-08-07
fix(fleet): 2026-08-07 Sigma spec-write contract break + domo gate-1 parity 24.6% → 71.7%

twells89:
## Why this PR grew

It started as domo gate-1 parity work. Mid-session **Sigma changed the workbook
code-rep write contract**, and PUTs that had succeeded an hour earlier began
returning 400. That breaks every converter in the fleet, not just domo, so the
fix landed here.

Proof it was the API and not us: `GET` the live spec and `PUT` it back
**unmodified** → 200; our built spec → 400.

## The new contract (live-verified)

| # | Requirement | Failure if violated |
|---|---|---|
| 1 | `document.elements` — one flat array | `"document.pages[].elements is no longer supported"` |
| 2 | `document.pages[]` = `{id, name}` only | as above |
| 3 | **every** element placed in the layout | `"element 'master' is not placed in layout"` |
| 4 | `<LayoutElement>`→`<Element>`, `<GridContainer>`→`<Container>` | opaque `incident-id` error |

(1) and (3) interact: hoisting **destroys** the page↔element association, so the
layout XML becomes the only record of which page an element is on. A hoist that
doesn't also guarantee a layout doesn't merely fail validation — it loses
information. That's why they're one method.

### The failure mode that cost the most time

An unknown layout tag returns
`{"message":"An error has occurred. Please try again later (incident-id=…)"}`.
It reads like a Sigma outage; it's your XML. Bisecting a known-good live
document against ours one part at a time (`live + my elements` → 200,
`live + my layout` → 400) localised it in two calls.

## The dangerous half was the gates, not the emitters

Both the layout lint and gate 4/7 **count placements and report the bad ones** —
so a parser that matches nothing reports nothing and the artifact reads CLEAN.

- `layout_lint.rb` parsed only old tag names **and** read elements only from
  `pages[].elements`. Against a current spec it found zero elements, leaving
  `el_kind` empty, which disarmed checks (a),(b),(d),(e),(f). Silent weakening
  that looks exactly like success.
- `assert-phase6-ran.rb` counted `<LayoutElement>` → 0 on a new-shape layout, a
  phantom gate-4/7 failure. The container check has the **opposite** polarity
  (it *exempts* banded pages from the auto-stack heuristic), so missing the new
  spelling produces phantom failures instead.

Proven fail-first per the standing rule: the three new lint tests **FAIL (3
FAILED)** against the pre-fix lint and pass after — they test the disarmament,
not the happy path.

## Two bugs caught by assertion rather than by a test failing

- **Silent duplication.** The first read-path version left the flat array in
  place while also populating pages, so `wrap()` counted every element twice —
  102 → 204. That's the whole workbook duplicated on the next save.
- **`<Container>` child-dropping.** `lib/layout.rb` warns that a wrong container
  tag *silently drops children*, so the rename was verified by readback rather
  than trusted on a 200: sent 26 containers / 66 children, read back 26 / 66,
  structurally identical.

## Live proof

The run now clears **post-and-readback → build-dashboard-layout → put-layout →
layout-2d-flag → render-visual → coverage-census** — every phase that touches the
spec. It stops at the warehouse-freshness guard (exit 9: Domo holding data 1 day
newer than the warehouse), which is that guard working correctly and is a data
staleness issue, not a contract one.

## Also in this PR (the original domo work)

Gate-1 parity **24.6% → 71.7%**, each step a measured cause: stale warehouse
copy, headers-only exports scored silently, month/week/quarter rendering,
rolling windows one period too wide (14 tiles), grouped limit-1 summary =
latest bucket. Plus a filter-operand fix: `filterType` is the literal `"LEGACY"`
on every Domo card filter, so the real `operand` was discarded — measured across
20 filters, **7 were emitted as the wrong filter entirely** (3 `NOT_IN` as their
exact inverse, 4 `GREATER_THAN` as a list include of `"0"`), which is exactly the
set of blank tiles previously filed as a defect.

## Verification

- domo suite green · all 16 shared suites green · tableau green except one
  pre-existing unrelated failure (`test-calc-discovery`) · powerbi unchanged
- `check-shared` clean across **681** copies
- `CodeRep.wrap` output PUT to the live API → **200**


twells89:
Superseded and salvaged. The fleet-wide layout-contract migration here (81 of 86 files) is already on main via #663 plus today's twelve per-converter PRs (#665-#676). The unique domo gate-1 parity oracle was lifted onto current main in #679, which also fixes a pre-existing failure in test-parity-oracle-dupenames.rb. Branch preserved.

PR #697, opened 2026-08-09
Revert "fix(tableau-to-sigma): quote emitted FIXED-LOD/window-helper SQL aliases"

twells89:
This reverts the alias-quoting change. A git bisect over the 2026-07-13
.. today window (known-good proven by a prior session reaching Phase 6
parity) names this commit as the first bad one: it breaks the data-model
POST for a workbook that previously migrated end to end.

Eight commits tested good in that bisect, including the shared code-rep
and document-envelope migrations and the converter migration. Only this
one tested bad.

Cause: quoting changes case-folding semantics. An unquoted alias is
folded to upper case by the warehouse; a quoted one keeps its literal
form, so any downstream reference expecting the folded name desyncs.
The original change verified producer/consumer agreement on the workbook
it was written for, but that workbook does not exercise the desync.

Reverting to restore a known-good spine. The underlying hyphenated-alias
bug is real and stays open; the re-fix must prove agreement on both the
workbook that motivated it and the one this broke.

tableau suite 262/263 (test-calc-discovery.rb needs a token plus a local
fixture, environmental); corpus 27/27; hygiene sweep clean.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


twells89:
Closing without merging — the premise was wrong.

A bisect named #695 as the first bad commit, but that bisect agent died mid-run and I acted on a partial log without validating what its bad-test actually measured. A migration run on 1.9.10 (which INCLUDES #695) has since completed successfully end to end: a fresh workbook (15 elements — 5 KPIs, line chart, 3 bar charts, table, 4 containers; layout tags Container/Element/Page) and a fresh 9-element data model, with rendered values matching Tableau.

So #695 is not categorically broken and main should not be reverted. The bad-test was most likely tripping on an unrelated credentials 400 or on the false-positive reference gate now filed separately, not on a real DM-POST failure.

Leaving main at 1.9.10. The hyphenated-alias fix stays in.

PR #715, opened 2026-08-13
feat(sigma-cross-org-port): Sigma org→org workbook port on the contents endpoints

twells89:
Adds `sigma-authoring/skills/sigma-cross-org-port` — copy a Sigma workbook into a
different Sigma org when both orgs already read the same warehouse tables
(demo/template/sample-data workbooks moving out of a shared org) — and cuts the
skill over to the current workbook **contents** endpoints.

## What is the purpose of this PR?

The skill remaps the org-scoped references in a workbook spec (connections, image
uploads), fixes the GET→POST asymmetries, creates the workbook in the target org,
and verifies structure + compilation + render. It is not a converter and not a
data migration — the target must already have the tables.

The latest change is an endpoint cutover. The skill was written against the legacy
`/v2/workbooks/spec` family, where the document sits under a `document` wrapper and
YAML bodies are accepted. Sigma's current, recommended surface reads
[`GET /v2/workbooks/{id}?includeContents=true`](https://github.com/twells89/sigma-migration-skills/blob/74958468cbfddcc0837b28b27344f325fd013cef/plugins/sigma-authoring/skills/sigma-cross-org-port/SKILL.md#L36-L51), creates via
`POST /v2/workbooks` with the doc under `contents`, updates via
`PUT /v2/workbooks/{id}/contents`, and validates via `POST /v2/workbooks` with
`dryRun: true`. These endpoints are **JSON only**. The scripts now write that
surface while still *reading* either envelope, so old saved specs keep working;
the legacy `/spec*` family also remains supported upstream.

## What should reviewers focus on?

The write path is the substance. [`port_workbook.py`](https://github.com/twells89/sigma-migration-skills/blob/74958468cbfddcc0837b28b27344f325fd013cef/plugins/sigma-authoring/skills/sigma-cross-org-port/scripts/port_workbook.py#L347-L356)
now emits a JSON create body keyed `contents` (with an optional `--dry-run`), and
[the read side](https://github.com/twells89/sigma-migration-skills/blob/74958468cbfddcc0837b28b27344f325fd013cef/plugins/sigma-authoring/skills/sigma-cross-org-port/scripts/port_workbook.py#L297-L303) of all three scripts accepts
a `contents` envelope, a legacy `document` envelope, or a bare document.

Please sanity-check the **dryRun semantics** baked into the Phase 5 examples and the
[envelope reference](https://github.com/twells89/sigma-migration-skills/blob/74958468cbfddcc0837b28b27344f325fd013cef/plugins/sigma-authoring/skills/sigma-cross-org-port/refs/spec-asymmetries.md#L10-L39): a clean dryRun returns
**HTTP 200 with the `valid` key omitted**, while a failing one returns **HTTP 400
with `{valid:false, errors:[…]}`** — so the [curl example](https://github.com/twells89/sigma-migration-skills/blob/74958468cbfddcc0837b28b27344f325fd013cef/plugins/sigma-authoring/skills/sigma-cross-org-port/SKILL.md#L309-L330)
tests the response body, not the status code. dryRun resolves dependencies, so it
catches `Dependency not found` that the old shape-only `/spec/verify` passed; it does
**not** replace the live compile probe, which is still required for the
`'Unknown column "[X]"'` string-literal class. `verify_port.py` also
[gains the native chart kinds](https://github.com/twells89/sigma-migration-skills/blob/74958468cbfddcc0837b28b27344f325fd013cef/plugins/sigma-authoring/skills/sigma-cross-org-port/scripts/verify_port.py#L37-L42) shipped since the
skill was written (treemap/sankey/funnel/gauge/box/waterfall/area) so its compile
probe covers them.

## How is this PR tested?

The cutover was exercised live against `tj-wells-1989` (`aws-api`). The new
`includeContents` GET returns the doc under `contents`; a create body emitted by the
rewritten `port_workbook.py` is accepted by `POST /v2/workbooks` — a workbook with
real dependency errors dry-ran to HTTP 400 `{valid:false, errors:[3]}`, a clean
round-trip dry-ran to HTTP 200 with `valid` omitted, and both persisted **0**
workbooks. `verify_port.py compile` reads a live workbook through the new GET and
probed its data elements clean; `structure` and `recover_images.py plan` both load
the `contents` envelope. All three scripts pass `python3 -m ast`, and the full
pre-commit governance hook (shared-file check, byte budgets, `hygiene-sweep.sh`,
portability lints) is green.

<details><summary>details — original end-to-end validation (unchanged by this cutover)</summary>

The skill's original build was validated against a live 92-element, 12-page
workbook (9 tables, 6 input tables, 29 buttons, 8 image uploads, 3 modals, 1 point
map) across two clouds/regions: 92/92 element ids preserved, layout placement /
settings / pages / panels / overlays identical, 17/18 data elements compile clean —
the 18th failing *identically* in the source org (a pre-existing `[Missing node/…]`
ref), which the bundled `baseline` probe demonstrates rather than asserts. That
two-org run was **not** re-executed this session (source-org credentials were not
available here); the same-org live checks above cover the endpoint cutover itself.

Findings encoded and documented by the skill:

| Finding | Consequence |
|---|---|
| `inode-<id>/COLUMN` column ids are **opaque** — never resolved against target inodes | Do NOT rewrite them (probe-verified) |
| `groupingId: "base"` emitted by GET, rejected by POST | strip it; schema says omit for ungrouped |
| Image upload keys are org-scoped; the API cannot create an upload | recover bytes from a PDF export, re-host as `kind: url` data URI |
| Overlay/modal pages absent from whole-workbook export | export each by `pageId` — that works |
| Create-validation rejects dangling action→column refs the source org tolerates | explicit `--map-column`, never inferred |
| API expands an undersized `<Container>` row span but **not its siblings** | full-height element silently collapses to a strip; porter pre-applies it |
| Input-table **data** has no automated path | structure ports, rows do not — reported, not hidden |

</details>

🏗️ Generated for Sigma by Claude Code

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
     {
       "name": "sigma-authoring",
       "source": "./plugins/sigma-authoring",
-      "description": "Canonical Sigma authoring skills every converter defers to: sigma-workbooks (workbook spec \u2014 element kinds, source kinds, controls, formulas, formatting, layout), sigma-data-models, and custom-sql-to-data-model. Install this ALONGSIDE any converter \u2014 they assume it is present for the current Sigma spec surface.",
+      "description": "Canonical Sigma authoring skills every converter defers to: sigma-workbooks (workbook spec \u2014 element kinds, source kinds, controls, formulas, formatting, layout), sigma-data-models, custom-sql-to-data-model, and sigma-cross-org-port (Sigma org \u2192 Sigma org workbook copies). Install this ALONGSIDE any converter \u2014 they assume it is present for the current Sigma spec surface.",
       "category": "authoring",
       "keywords": [
         "sigma",
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,6 +63,7 @@ Maturity labels (`gold` / `live` / `foundation` / `scaffold`) are defined in
 | Convert a Mode Report (SQL Queries + Charts) → Sigma | `mode-to-sigma` | foundation | `plugins/mode-to-sigma/skills/mode-to-sigma/` |
 | Author / repair Sigma workbooks (canonical spec) | `sigma-workbooks` | live | `plugins/sigma-authoring/skills/sigma-workbooks/` |
 | Author / repair Sigma data models (canonical spec) | `sigma-data-models` | live | `plugins/sigma-authoring/skills/sigma-data-models/` |
+| Copy a Sigma workbook into another Sigma org that already has the same data | `sigma-cross-org-port` | live | `plugins/sigma-authoring/skills/sigma-cross-org-port/` |
 
 Assessments are read-only (never write to the source or post to Sigma); run one
 to pick what to convert, then hand off to the matching converter.
--- a/plugins/sigma-authoring/.claude-plugin/plugin.json
+++ b/plugins/sigma-authoring/.claude-plugin/plugin.json
@@ -2,8 +2,8 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "sigma-authoring",
   "displayName": "Sigma Authoring (Workbooks + Data Models)",
-  "version": "1.12.3",
-  "description": "The canonical Sigma authoring skills every migration converter defers to for the current spec surface: sigma-api (authentication), sigma-workbooks (workbook spec \u2014 element kinds, source kinds, controls, formulas, formatting, layout), sigma-data-models (data model authoring), and custom-sql-to-data-model. Install this alongside any converter \u2014 the converters assume it is present.",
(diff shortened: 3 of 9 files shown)


twells89:
Closing unmerged. A Sigma→Sigma cross-org workbook copy is not a BI-tool
migration, so it doesn't belong alongside the converters — same call as
`twells89/alteryx-to-sigma`.

The skill now lives in its own repo, cut over to the current workbook **contents**
endpoints: **https://github.com/twells89/sigma-cross-org-port**

Nothing here landed on `main` (this PR was never merged), so there's nothing to
revert. The branch `feat/sigma-cross-org-port` is left in place (not deleted) — it
carries the full history including the `/contents` cutover (`74958468`), in case
it's useful to revisit.

Issue #752, opened 2026-08-21
corpus/tableau/preagg-kpi fails under a non-UTF-8 locale — four scripts missing the lib/cli_encoding require


CurtisDeCastro:
## Summary

`corpus/tableau/preagg-kpi` fails on `main` (`cf5e572`) when the shell locale is
unset, and passes when it is UTF-8. The corpus itself is fine — the failure is
environmental, which is probably why it isn't showing up for everyone.

```
LANG unset          ->  corpus: 11/12 cases pass   (preagg-kpi FAILs)
LANG=en_US.UTF-8    ->  corpus: 12/12 cases pass
```

## Root cause

Four scripts on this case's path do **not** `require_relative 'lib/cli_encoding'`:

| script | requires cli_encoding |
|---|---|
| `audit-agg-semantics.rb` | no |
| `assert-phase6-ran.rb` | no |
| `audit-lod-calcs.rb` | no |
| `parse-twb-layout.rb` | no |

`lib/cli_encoding.rb` exists specifically to prevent this class of failure — its
own header documents it as "Field failure, BOTH field-workbook runs (3 crashes
total): under an unset/C locale, Ruby 2.6 tags ARGV strings ASCII-8BIT; the
first em-dash … raises `Encoding::UndefinedConversionError` inside
`JSON.generate`."

The `preagg-kpi` fixtures contain em-dashes (`workbook-content.twb`,
`agg-semantics.entries.json`, `dm-spec.fixture.json`, `checks.sh`,
`MANIFEST.md`), so the case walks straight into it.

Observed with `Encoding.default_external == US-ASCII` (system Ruby 2.6.10,
macOS):

```
assert-phase6-ran.rb:3158:in `scan': invalid byte sequence in US-ASCII (ArgumentError)

audit-agg-semantics.rb:90:in `<main>':
  "\xE2" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
  from json/common.rb:286:in `pretty_generate'

-e:8: invalid multibyte char (US-ASCII)      # the inline `ruby -rjson -e` assertions in checks.sh
```

## Reproduce

```bash
cd corpus
env -u LANG -u LC_ALL bash run-corpus.sh --check tableau   # 11/12, preagg-kpi FAILs
LANG=en_US.UTF-8 bash run-corpus.sh --check tableau        # 12/12
```

## Suggested fix

Add `require_relative 'lib/cli_encoding'` to the four scripts above (and audit
[...]

PR #762, opened 2026-08-25
feat: .rb/.py twin API-parity gate (#753)

### LINKED ISSUE #753: shared/lib Python twins silently lag their Ruby counterparts (sigma_rest, code_rep, metric_binding)
## Summary

Several `shared/lib` modules are maintained as `.rb` + `.py` twins, but some of
the Python twins are missing functions their Ruby counterparts export. They
import cleanly, so nothing surfaces the gap — a caller ported to Python just
hits `AttributeError` at runtime, or silently takes a different path.

Verified against `main` (`cf5e572`):

| module | in `.rb`, missing from `.py` | Ruby callers outside the module |
|---|---|---|
| `sigma_rest` | `list_entries` | `assert-phase6-ran.rb`, `discover-columns.rb`, `discover-warehouse-columns.rb` |
| `sigma_rest` | `token_stale?` | `migrate-tableau.rb` |
| `code_rep` | `flatten_elements` | `column_census.rb` |
| `code_rep` | `canonicalize_document`, `fold_legacy_theme` | none found |
| `metric_binding` | `dedup_by_name`, `walk_chain` | none found |

The `sigma_rest` two are the ones that bite: both have real callers on the
migrate spine.

Some of these may be intentional (internal helpers that the Python side
inlines, or functions genuinely not needed there yet) — in which case a comment
saying so would keep the next person from re-reporting it.

## Why it matters

The `.py`/`.rb` pairing reads as "these are equivalent," and that's the
assumption anyone porting a caller will make. Right now equivalence is asserted
by file naming alone; nothing tests it. `tools/check-shared.rb` enforces
byte-identical *vendoring* of a file into each plugin, but nothing checks that a
`.py` twin actually matches its `.rb`.

## Suggested guard

Two options, cheap to expensive:

1. **Symbol-parity check in CI** — for every `shared/lib/<m>.{rb,py}` pair,
   diff the exported function names and fail on a gap that isn't explicitly
   allow-listed. Catches exactly the table above.
2. **Behavioural differential** — drive both implementations with the same
   inputs and compare outputs. I've been doing this for a Python port of the
   Tableau skill: a generic pair of drivers (`Module.send(fn, *args)` vs
   `getattr(mod, fn)(*args)`) over a JSON case file, so adding a module is just
   adding cases. Two wrinkles worth knowing if you go this route — Ruby
   predicates end in `?` and Python names don't, and exception *class names*
   differ across languages, so "both raised" has to count as a match.

Happy to contribute either if useful.

## Reproduce

```bash
for m in sigma_rest code_rep metric_binding; do
  echo "== $m"
  comm -13 \
    <(grep -oE '^def [a-z_0-9]+' shared/lib/$m.py | sed 's/def //' | sort -u) \
    <(grep -oE '^\s*def [a-z_0-9?!]+' shared/lib/$m.rb | sed 's/.*def //;s/[?!]$//' | sort -u)
done
```

twells89:
Closes #753.

The premise is exactly right: the `.rb`/`.py` pairing in `shared/lib` reads as "these are equivalent," `check-shared.rb` only enforces byte-identical *vendoring* of one file into each plugin, and nothing checked that a `.py` twin still exports what its `.rb` does. Building the guard was worth it.

Verifying the table changed the design, though, so the specifics are worth recording.

### Five of the nine rows are false positives

The `grep '^def '` diff doesn't model how each language marks something internal:

| Reported missing | Actually |
|---|---|
| `metric_binding.dedup_by_name` | `_dedup_by_name` (`.py:151`) |
| `metric_binding.walk_chain` | `_walk_chain` (`.py:122`) |
| `sigma_rest.token_stale?` | `_token_stale` (`.py:190`) |
| `code_rep.flatten_elements` | `_flatten_elements` (`.py:150`) |
| `code_rep.fold_legacy_theme` | `_fold_legacy_theme` (`.py:31`) |

Python's leading underscore *is* the privacy marker. And on the Ruby side, all three flagged `code_rep` methods — including `canonicalize_document` — sit below that module's `private` on line 145, so none of them was ever public API. **`code_rep` has zero public gaps.**

### Both "the ones that bite" claims dissolve

`list_entries`' three cited callers are **comments documenting deliberate non-use**:

```
assert-phase6-ran.rb:1474   # prevent. Local loop rather than Sigma.list_entries: this gate deliberately
discover-columns.rb:131     #    `page` shape list_entries assumes — list_entries silently stops after page
discover-warehouse-columns.rb:31  # `nextPageToken`/expects `pageToken`, not list_entries' assumed `nextPage`/
```

Those endpoints use `nextPageToken`/`pageToken`, so they hand-roll the loop on purpose. The one real caller — `fidelity-loop.rb:558` — isn't in the list. Worth knowing before porting it: a blind translation would inherit the wrong pagination contract.

And `code_rep.flatten_elements`' cited caller, `column_census.rb`, **defines its own** `flatten_elements` at line 147 and never requires `code_rep`.

### What the report missed — the biggest real divergence

`tableau_rest` exports **eight** functions with no Python counterpart, including the `query_datasource` / `view_data_filtered` data-fetch spine. That's larger than anything in the table, and squarely on the path of a Python port of the Tableau skill.

Allow-listed with reasons rather than ported: the data-fetch functions need live-Tableau validation of pagination and filter semantics, and a blind translation would be worse than the gap. The gate makes it visible instead of silent. Happy to take the port as a follow-up.

## The gate

Your option 1, with the normalization that makes it usable. Modelling privacy properly was the difference between 12 phantom findings and 0 — and a gate that cries wolf gets switched off, after which the real gap ships.

It handles: Ruby `private` regions (with `def self.` escaping them, which `coverage_catalog` relies on), `module_function`, `?`/`!` suffixes, `initialize` ↔ `__init__`, class methods on both sides, declared renames (`warehouse_transforms` `apply` → `apply_transforms`), and Python's `_` prefix.

Three outcomes rather than one:

- **MISSING** — no counterpart at all → fails unless allow-listed with a reason
- **VISIBILITY** — exists as `_name` → reported, doesn't fail. This is the class the `^def` diff mistook for missing, and it's still the actionable fact for a porter: it's there, use the underscored name.
- **ALLOWED** — deliberate non-port → reported with its reason

A **stale** allow-list entry also fails, so the exemption list can't quietly become permanent.

On `main`:

```
OK: 14 .rb/.py twin pair(s) reconciled — no missing function.
    (9 documented non-port, 12 visibility-only difference.)
```

You flagged the `?`-suffix and cross-language exception-name wrinkles for the behavioural approach — both real. I've skipped option 2 here: the symbol gate catches this class for a fraction of the cost, and your differential harness is better owned by the port that needs it.

## Also corrected

`sigma_rest.py` and `tableau_rest.py` both claimed parity was "enforced by ... the cross-impl parity harness." No such harness exists — the phrase appears nowhere but in those two docstrings — and `test_{sigma,tableau}_rest.py` never invoke Ruby; they hand-encode the Ruby behaviour in Python. They now state what actually holds, and record the deliberate non-ports so this isn't re-reported.

## Verification

Gate proven to **fail** before being trusted. `tools/test-lint-twin-parity.rb` (17 checks) asserts a deleted function fails and is named, that porting it clears the gate, that a stale allow-list entry fails, and that all seven false-positive shapes stay quiet. Also verified against the real tree by deleting `_walk_chain` from `metric_binding.py` — fails, names `walk_chain`, passes again on restore.

`check-shared.rb` clean (815 copies), governance suite exit 0, both Python twin test suites pass, versions bumped. Wired into CI and the governance hook.

### DIFF
--- a/.githooks/run-governance-checks.sh
+++ b/.githooks/run-governance-checks.sh
@@ -34,6 +34,7 @@ fi
 [ -f tools/check-agent-variants.rb ] && { ruby tools/check-agent-variants.rb || fail=1; }
 [ -f tools/lint-skill-paths.rb ] && { ruby tools/lint-skill-paths.rb || fail=1; }
 [ -f tools/lint-uplift-contracts.rb ] && { ruby tools/lint-uplift-contracts.rb || fail=1; }
+[ -f tools/lint-twin-parity.rb ] && { ruby tools/lint-twin-parity.rb || fail=1; }
 [ -f tools/check-cognos-bundle.rb ] && { ruby tools/check-cognos-bundle.rb || fail=1; }
 # bootstrap<->doctor KEEP-IN-LOCKSTEP guard (E2.1): bootstrap.sh duplicates
 # doctor.sh's probes (py_real, the vm-node candidate globs, Test-RealPython)
@@ -56,6 +57,7 @@ if [ "${CI:-}" = "true" ]; then
   [ -f tools/test-lint-skill-paths.sh ] && { bash tools/test-lint-skill-paths.sh || fail=1; }
   [ -f tools/test-lint-uplift-contracts.rb ] && { ruby tools/test-lint-uplift-contracts.rb || fail=1; }
   [ -f tools/test-lint-twb-encoding.rb ] && { ruby tools/test-lint-twb-encoding.rb || fail=1; }
+  [ -f tools/test-lint-twin-parity.rb ] && { ruby tools/test-lint-twin-parity.rb || fail=1; }
 fi
 if [ "$fail" -ne 0 ]; then
   echo "" >&2
--- a/.github/workflows/corpus-check.yml
+++ b/.github/workflows/corpus-check.yml
@@ -429,6 +429,18 @@ jobs:
         # (raw dm-spec.json read + .scan) in a throwaway tree and asserts exit 1,
         # plus the safe shapes that must stay quiet.
         run: ruby tools/test-lint-twb-encoding.rb
+      - name: Lint — .rb/.py twin API parity (#753)
+        # The .rb/.py pairing in shared/lib reads as "these are equivalent", but
+        # nothing checked it: check-shared.rb enforces byte-identical VENDORING of
+        # one file into each plugin, not that a .py twin exports what its .rb does.
+        # A ported caller just hits AttributeError at runtime.
+        run: ruby tools/lint-twin-parity.rb
+      - name: Self-test — the twin-parity gate must FAIL on a deleted function
+        # And must stay QUIET on the seven shapes a naive `grep '^def '` diff gets
+        # wrong (Python's `_` privacy marker, ruby `private` regions, `def self.`,
+        # `?`/`!` suffixes, initialize/__init__, declared renames). Five of the nine
+        # rows in the original report were one of those.
+        run: ruby tools/test-lint-twin-parity.rb
       - name: Run offline test_*.py suites
         # EXPLICIT allow-list — creds-free, network-free Python tests only.
         # PyYAML is needed by the looker offline dashboard parser
--- a/plugins/gooddata-to-sigma/.claude-plugin/plugin.json
+++ b/plugins/gooddata-to-sigma/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "gooddata-to-sigma",
   "displayName": "GoodData \u2192 Sigma",
-  "version": "1.2.19",
+  "version": "1.2.20",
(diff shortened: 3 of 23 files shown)


twells89:
**Merge order:** land #761 (the #752 crash fix) first — it's the urgent one and it's already green.

These two touch the same wiring files, so I've pre-emptively de-collided the version bumps here (this PR now claims one patch above whatever #761 claims). That leaves only two files that can conflict after #761 merges, both trivial appends:

- `.githooks/run-governance-checks.sh` — #761 adds `test-lint-twb-encoding.rb`, this adds `lint-twin-parity.rb` + its self-test
- `.github/workflows/corpus-check.yml` — both append a step near the same anchor

Ping me when #761 lands and I'll rebase.

PR #789, opened 2026-09-14
fix(looker-to-sigma): fan-out-safe converter and input-completeness preflight

twells89:
## What is the purpose of this PR?

A real LookML project was assessed through this skill and reported back as essentially unconvertible. It is not, and the way we reached that conclusion is worth fixing permanently.

The export we were given was partial: the explore referenced 149 distinct physical views and 38 were supplied. Every unexported joined view becomes a `LOOKER_SCRATCH.<VIEW>` placeholder carrying its own loud warning, so the run emitted hundreds of warnings and read as "this tool cannot convert our model." A missing-input problem was reported as a capability problem, and nothing in the skill distinguished the two.

Two documentation defects made that reading more likely. `SKILL.md` directed operators to a `looker-assessment` sibling skill that does not exist in this repo, and it listed NDT (`explore_source`), PDT `datagroup`/`persist_for`, and `many_to_many` as "not yet converted" when all three have been handled for months — verified by reading `resolveNdtToSql`, `PERSIST_HINTS`, and the M:N mapping in the converter itself rather than assuming.

The vendored converter is refreshed in the same change, which is what makes the behaviour fixes reach an operator: the skill runs `converter/lookml.mjs` by default, so nothing in the conversion path changes until that bundle is regenerated here.

## What should reviewers focus on?

[`check_input_completeness.py`](https://github.com/twells89/sigma-migration-skills/blob/9478977987991bef02fd0a06ecf050c01ccfeb3d/plugins/looker-to-sigma/skills/looker-to-sigma/scripts/check_input_completeness.py) is new and runs as Phase 0b.1, before conversion. It resolves `from:` aliasing first so N aliases of one view count once, then reports physical views referenced versus supplied, joins wirable versus not, whether any explore exists at all, and whether the target explore is `extension: required` — an abstract base nobody runs directly, whose concrete extending explores live elsewhere. On a complete project it prints one all-clear line and exits 0; it never blocks, because a partial export is still worth converting as a deliberate slice. Comment stripping is quote-aware so a commented-out `join:` cannot inflate the counts, and colour literals like `'#fff'` survive.

On the project that prompted this it reports 38 views provided, 149 referenced, 124 missing, 100 of 318 joins (31%) wirable, and flags the abstract explore.

The wording in [Phase 0b](https://github.com/twells89/sigma-migration-skills/blob/9478977987991bef02fd0a06ecf050c01ccfeb3d/plugins/looker-to-sigma/skills/looker-to-sigma/SKILL.md#L340-L375) is the part worth reading closely, since it is the actual fix for the wrong verdict: it states plainly that an incomplete input must not be reported as "cannot convert." Please confirm that framing matches how you want operators to escalate. The same section replaces the non-existent `looker-assessment` pointer with concrete guidance, including using Looker System Activity to scope by what is actually queried — on a large estate that single input is usually the difference between porting a few hundred fields and a few thousand.

## What the refreshed converter changes

A `sql_on` that ANDs several `${a.b} = ${c.d}` pairs now produces one relationship carrying every key pair. Previously only the first pair was captured, and the "unparsable sql_on" warning fired only when nothing matched — so a composite join silently became a single-key relationship that still POSTs, still queries, and fans out. In the project above, 288 of 318 joins (91%) are composite and 172 are 4-key. Liquid `{% condition %}` blocks are stripped before key extraction so their inner text is not read as a key, and literal predicates (`${view.col} = 5`) are reported rather than dropped, since a relationship carries no predicate.

`primary_key: yes` now becomes element `uniqueKeys`, which is what Sigma's semantic-aggregates compiler uses to track a table's grain and aggregate to it before display. A view with no `primary_key` warns that its grain is undeclared.

The [coverage section](https://github.com/twells89/sigma-migration-skills/blob/9478977987991bef02fd0a06ecf050c01ccfeb3d/plugins/looker-to-sigma/skills/looker-to-sigma/SKILL.md#L596-L620) documents both, including why `sql_distinct_key` is deliberately *not* mapped into `uniqueKeys`: it is a measure-level de-dup grain that routinely spans joined views, whereas `uniqueKeys` lists columns on the element itself. Declaring a cross-view grain there would be the wrong grain, and a wrong grain under semantic aggregates produces wrong numbers — the failure this metadata exists to prevent.

<details><summary>details — converter/PROVENANCE.json local_patches cleared</summary>

`converter/PROVENANCE.json` recorded three fixes applied directly to the vendored bundle that had never landed in the converter source, each with an explicit warning that refreshing the bundle before they land would drop them. The first push of this branch did exactly that and the corpus caught it: 35/35 → 34/35, with `dynamicParameters` going from 3 to 0 on `skilltest-course-performance`.

Rather than re-patch the bundle and preserve the drift, the three were landed in the converter source (Looker filter normalization, `lookCollectDynamicParameters` and its return wiring, and the `${TABLE}` allowance in branch validation). The bundle is now regenerable from source with no drift and `local_patches` is cleared — reviewers should confirm that clearing is correct rather than premature.

</details>

Plugin version moves 1.2.33 → 1.3.0: minor rather than patch, for the new preflight script alongside converter behaviour changes, with no breaking API.

## How is this PR tested?

The artifact this PR ships is the vendored bundle, so it was exercised directly rather than by proxy. A fixture combining a two-key composite join with a Liquid `{% condition %}` in its `sql_on`, a composite `primary_key`, a `derived_table`, and a `parameter:` with allowed values was converted by this branch's `converter/lookml.mjs` and POSTed to a live Sigma org on a Snowflake connection.

`uniqueKeys` persisted with both columns and `relationshipType: "N:1"` persisted with two key pairs through POST → GET. All 14 columns resolved to concrete types with zero error-typed. The generated SQL joins on both keys (`on Q3.CUSTOMER_KEY_32 = Q4.CUSTOMER_KEY and Q3.ORDER_DATE_KEY_33 = Q4.ORDER_DATE_KEY`), and the query returned real warehouse rows with the related column resolved through the relationship rather than NULL. `dynamicParameters` reported `metric_picker` as deterministic, confirming the formerly bundle-local patch now works from source. The test data model was deleted and the target folder confirmed clean.

<details><summary>details — repo gates and offline suites</summary>

Corpus 35/35, with the baseline established on `origin/main` in a separate worktree first, since some corpus cases can already be failing and a green local run means nothing without it.

All 13 looker offline Python tests and both looker Ruby tests from the CI allow-list pass, including `test_filter_normalization.py` and `test_dynamic_parameters.py`, which are the two that specifically cover the formerly bundle-local patches.

`tools/check-converter-provenance.sh` and `tools/check-plugin-version-bump.sh` both pass against `origin/main..HEAD`. Worth noting for anyone reproducing: the version-bump gate passes vacuously on uncommitted work ("no plugin directories changed"), so it must be run after committing.

</details>

Closes `beads-sigma-yqav`.

🏗️ Generated for Sigma by Claude Code


twells89:
## Live org validation of the shipped artifact

Flagging this explicitly because the earlier verification in the PR body exercised the **converter source**, not the **vendored bundle this PR actually ships** — and after the re-vendor at `16ad063` those are no longer trivially the same thing. Re-ran end-to-end against a live Sigma org + Snowflake connection, driving `converter/lookml.mjs` from this branch directly.

Fixture deliberately exercises all three changed areas at once: a **2-key composite join** with a Liquid `{% condition %}` block in the `sql_on`, a **composite `primary_key`** (`ORDER_ID` + `ORDER_LINE`), a `derived_table`, and a **`parameter:` with allowed values** (to cover the newly-ported `dynamicParameters` patch).

**Converter output**
- 1 relationship carrying **2 key pairs** (not 2 one-key relationships)
- `uniqueKeys: [ORDER_ID, ORDER_LINE]` on the fact element
- `dynamicParameters: [{name: metric_picker, deterministic: true}]` — the ported patch working from source, not a bundle-local patch

**POST → GET round-trip**
- `uniqueKeys` persisted, both columns
- `relationshipType: "N:1"` persisted, `keys.length == 2`

**Compile + query**
- 14 columns, **0 error-typed**
- Generated SQL joins on **both** keys:
  `... on Q3.CUSTOMER_KEY_32 = Q4.CUSTOMER_KEY and Q3.ORDER_DATE_KEY_33 = Q4.ORDER_DATE_KEY`
- Query returns real warehouse rows, with the related column resolved through the relationship rather than NULL

**Cleanup:** test data model deleted; target folder confirmed empty of test artifacts.

Combined with corpus 35/35 and the 13 looker offline Python + 2 Ruby suites from the CI allow-list, the artifact in this PR is verified end-to-end rather than only structurally.

